PR #26, opened 2019-11-26
Fixes for autoscaling API changes

kianjones4:
* Checks aws error message instead of error code when looking for instance not found exception for TerminateInstanceInAutoScalingGroup calls, since the error code is different from the code we were previously using in EC2

* Added throttling fix to retry when upgrade manager gets throttled by rate limiting.


Dockerfile
```diff
@@ -7,6 +7,7 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
+COPY pkg pkg
```
shrinandj: Why was this needed?

controllers/rollingupgrade_controller_test.go
```diff
@@ -321,7 +322,7 @@ func TestTerminateNodeErrorNotFound(t *testing.T) {
 	ruObj := &upgrademgrv1alpha1.RollingUpgrade{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"}}
 	rcRollingUpgrade := &RollingUpgradeReconciler{ClusterState: NewClusterState()}
 	mockAutoscalingGroup := MockAutoscalingGroup{errorFlag: true, awsErr: awserr.New("InvalidInstanceID.NotFound",
-		"some message",
+		"ValidationError: Instance Id not found - No managed instance found for instance ID i-0bba",
```
shrinandj: Nice!

pkg/log/log.go
```diff
@@ -0,0 +1,200 @@
+package log
```
shrinandj: Why is this needed?

kianjones4: This is so that we can log the retries if the rollup controller autoscalling API calls get throttled. Similar to the way it was implemented in lifecycle manager https://github.com/keikoproj/lifecycle-manager/blob/7661ea7bd6fc4f5ddd92243252500a8e634ebc57/cmd/serve.go#L156-L159

PR #35, opened 2019-12-09
wait for node unjoined

### LINKED ISSUE #33: Check node object is gone before proceeding
Feature Request

**What happened**:
Rolling upgrade waits a predefined amount of time after terminating an instance, but even if the node object is still around, it moves on to terminate the next instance

it would be nice if outside of waiting the post-terminate defined time, rollingupgrade also verified that the node object is gone before moving to the next instance.
This is very useful with components such as `lifecycle-manager` for cases where there is a delay with the termination of a node.

eytan-avisror:
Fixes #33 

This PR waits for node object to be unjoined before proceeding with rolling upgrade.

There is a hard limit on how long to wait which is now capped at an hour, if timeout occurs, upgrade will disregard and proceed anyway - we can make this configurable in CR if desired.

Added some unit test for new method.

## Testing

- Unit tests passing
- Manual testing - TBD in progress

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -76,6 +76,13 @@ const (
 	ShellBinary = "/bin/sh"
 )
 
+var (
+	// TerminationTimeoutSeconds is the timeout threshold for waiting for a node object unjoin
+	TerminationTimeoutSeconds = 3600
+	// TerminationSleepIntervalSeconds is the polling interval for checking if a node object is unjoined
+	TerminationSleepIntervalSeconds = 30
+)
+
 var DefaultRetryer = awsclient.DefaultRetryer{
 	NumMaxRetries:    250,
 	MinThrottleDelay: time.Second * 5,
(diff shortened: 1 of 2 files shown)


controllers/rollingupgrade_controller.go
```diff
@@ -761,6 +802,12 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
 		return
 	}
 
+	err = r.WaitForTermination(nodeName, r.generatedClient.CoreV1().Nodes())
```
shrinandj: Instead of getting all the nodes and filtering, can we simply Get the required node?

`r.generatedClient.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})` should do it

eytan-avisror: Yes, but in that case we will have to rely on NotFound error.
I will make the change

eytan-avisror: Done

Issue #30, opened 2019-12-03
Upgrade manager fails to actually upgrade nodes because of AWS token issues


shrinandj:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG REPORT

**What happened**:
Upgrade manager fails because it cannot get AWS credentials. It consistent fails with the following errors:
```
time="2019-12-03T05:09:43Z" level=info msg="retryable: RequestError: send request failed\ncaused by: Put http://169.254.169.254/latest/api/token: net/http: request canceled (Client.Timeout exceeded while awaiting headers) -- ec2metadata/GetToken, will retry after 2.898737352s"
time="2019-12-03T05:09:49Z" level=info msg="retryable: RequestError: send request failed\ncaused by: Put http://169.254.169.254/latest/api/token: net/http: request canceled (Client.Timeout exceeded while awaiting headers) -- ec2metadata/GetToken, will retry after 3.287156464s"
time="2019-12-03T05:09:51Z" level=info msg="retryable: RequestError: send request failed\ncaused by: Put http://169.254.169.254/latest/api/token: net/http: request canceled (Client.Timeout exceeded while awaiting headers) -- ec2metadata/GetToken, will retry after 3.542636997s"
time="2019-12-03T05:09:51Z" level=info msg="retryable: RequestError: send request failed\ncaused by: Put http://169.254.169.254/latest/api/token: net/http: request canceled (Client.Timeout exceeded while awaiting headers) -- ec2metadata/GetToken, will retry after 4.523316716s"
time="2019-12-03T05:09:52Z" level=info msg="retryable: RequestError: send request failed\ncaused by: Put http://169.254.169.254/latest/api/token: net/http: request canceled (Client.Timeout exceeded while awaiting headers) -- ec2metadata/GetToken, will retry after 4.209153918s"
time="2019-12-03T05:09:57Z" level=info msg="retryable: RequestError: send request failed\ncaused by: Put http://169.254.169.254/latest/api/token: net/http: request canceled (Client.Timeout exceeded while awaiting headers) -- ec2metadata/GetToken, will retry after 4.761618292s"
```

This is most likely because of https://github.com/aws/aws-sdk-go/issues/2972.

**What you expected to happen**:
There should be no errors and upgrade-manager should actually perform the upgrade.

**How to reproduce it (as minimally and precisely as possible)**:
Just update the launch-config of an IG and run upgrade-manager.

**Anything else we need to know?**:
[...]

PR #36, opened 2019-12-09
Add build badge

kianjones4:
Adds the Semaphore build badge to the README


pkg/log/go.mod
```diff
@@ -3,6 +3,6 @@ module github.com/keikoproj/upgrade-manager/pkg/log
 go 1.13
 
 require (
-	github.com/aws/aws-sdk-go v1.25.0
+	github.com/aws/aws-sdk-go v1.25.42
```
shrinandj: This was intentionally rolled back to v1.25.0 in https://github.com/keikoproj/upgrade-manager/issues/30. Newer versions of the SDK use an updated mechanism of getting the aws creds and require some changes to the AMI before doing that.

Has something changed?

PR #38, opened 2019-12-13
Parallel Reconcile Limit

### LINKED ISSUE #34: Ability to control max parallel upgrades
FEATURE REQUEST

**What happened**:
On some big clusters with many different IGs, upgrade-manager can start upgrades of all of them in parallel. This has performance benefits. However, at times, too many parallel upgrades can cause hitting AWS API limits. Or at least it can significantly increase the AWS API rate which can result in other components hitting rate limits.

It should be possible to control how many `RollingUpgrade` objects get processed in parallel. This would be a setting at the controller level. Admins can set the correct amount of parallelism depending on which and how many other components exist that can cause AWS API rate limits.

eytan-avisror:
Fixes #34 

This adds a bounded channel to block reconciles over N concurrent reconciles.

The use of controller flag `--max-parallel` allows to set a number of maximum concurrent reconciles, the default `0` indicates there is no max and disables this feature.

The bounded channel will block the reconcile from happening until a different goroutine releases it's lock, the channel size indicates how many parallel go routines can execute.

## Testing

Ran controller manually with 3 IGs and `--max-parallel` set to 2, validated scaling groups are handled 2 at a time.


controllers/rollingupgrade_controller.go
```diff
@@ -602,6 +622,13 @@ func (r *RollingUpgradeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 			logr.Info("Sync map with invalid entry for ", "name", ruObj.Name)
 		}
 	} else {
+
+		if r.maxParallel != nil {
+			logr.Info("trying to acquire semaphore lock...")
+			r.AcquireSemaphore()
```
shrinandj: Won't this block the main thread in the running controller when there are more than maxParallel  CRs?

eytan-avisror: Good point, I moved the semaphore lock inside the goroutine, now it should no longer block the main reconcile thread, but the actual `Process()` goroutine instead.

eytan-avisror: BTW, `Reconcile()` on itself is called async, so even if the blocking channel is inside `Reconcile()` it won't block new reconciles from happening (or atleast getting queued)/

eytan-avisror:
@shrinandj 
I've made the following changes:
- Update `controller-runtime` and `client-go` to latest versions, this is required in order to use controller's `MaxConcurrentReconcile` feature.
- Instead of using semaphore/channel to block thread, we now simply pass the flag to the controller
- Removed goroutine creation for `r.Process()`, now `Reconcile()` is the only goroutine, and number of concurrent threads is controlled by `--max-parallel` flag.

I've set the default to some high number e.g. 10, let me know if you think we should make this a larger number, but the number of concurrent reconciles can now be fully controlled via `--max-parallel`.

I've tested this and it works without the issues we've discussed.

controllers/rollingupgrade_controller.go
```diff
@@ -602,7 +610,7 @@ func (r *RollingUpgradeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 			logr.Info("Sync map with invalid entry for ", "name", ruObj.Name)
 		}
 	} else {
-		go r.Process(&ctx, ruObj)
+		r.Process(&ctx, ruObj)
```
eytan-avisror: calling `r.Process()` as a goroutine bypasses the native mechanism in controller-runtime for `MaxConcurrentReconciles`, it is now blocking and `Reconcile()` is called asynchronously according to worker count (`--max-parallel`)

controllers/rollingupgrade_controller.go
```diff
@@ -615,6 +623,7 @@ func (r *RollingUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.generatedClient = kubernetes.NewForConfigOrDie(mgr.GetConfig())
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&upgrademgrv1alpha1.RollingUpgrade{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: r.maxParallel}).
```
eytan-avisror: This adds the option `MaxConcurrentReconciles` when starting the controller according to what is passed in the flag.

controllers/rollingupgrade_controller.go
eytan-avisror: Only downside here is that whomever was used to `infinite` worker count as the behavior (due to calling `r.Process()` as a goroutine), is now getting limited to whatever default we set (in this case 10) as max parallel reconciles.

tekenstam:
Should we add documentation about the `max-parallel` flag (and the other flags) to README.md or some other doc page?

eytan-avisror:
> Should we add documentation about the `max-parallel` flag (and the other flags) to README.md or some other doc page?

We could, however I don't know if it should be in README.md. @shrinandj any preference on where we should document this, given you are OK with this implementation?

Issue #42, opened 2020-01-10
Support AWS LaunchTemplates


uthark:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

Upgrade Manager doesn't support LaunchTemplates as of now.
Do you have any plans on adding this feature (btw, kops supports LaunchTemplates)

Would you accept PR that adds support for it?

shrinandj:
> Would you accept PR that adds support for it?

Absolutely!

PR #43, opened 2020-01-13
[WIP] Add support for LaunchTemplates

uthark:
Hello!
Fixes https://github.com/keikoproj/upgrade-manager/issues/42

Adds support for LaunchTemplates and MixedInstancesPolicy.

I did manual functional tests, works as expected.

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -40,8 +40,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/go-logr/logr"
-	upgrademgrv1alpha1 "github.com/keikoproj/upgrade-manager/api/v1alpha1"
-	log "github.com/keikoproj/upgrade-manager/pkg/log"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	v1errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -51,6 +49,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	upgrademgrv1alpha1 "github.com/keikoproj/upgrade-manager/api/v1alpha1"
+	log "github.com/keikoproj/upgrade-manager/pkg/log"
 )
 
 const (
@@ -446,7 +447,8 @@ func (r *RollingUpgradeReconciler) runRestack(ctx *context.Context, ruObj *upgra
 	// set the state of instances in the ASG to new in the cluster store
 	r.ClusterState.initializeAsg(*asg.AutoScalingGroupName, asg.Instances)
 
-	currentLaunchConfigName := aws.StringValue(asg.LaunchConfigurationName)
+	launchDefinition := NewLaunchDefinition(asg)
+
 	processedInstances := 0
 
 	inProgress, err := r.getInProgressInstances(asg.Instances)
@@ -479,7 +481,7 @@ func (r *RollingUpgradeReconciler) runRestack(ctx *context.Context, ruObj *upgra
 		}
 
 		// update the instances
-		err := r.UpdateInstances(ctx, ruObj, instances, currentLaunchConfigName, KubeCtlCall)
+		err := r.UpdateInstances(ctx, ruObj, instances, launchDefinition, KubeCtlCall)
 		processedInstances += len(instances)
 		if err != nil {
 			return processedInstances, err
@@ -754,7 +756,7 @@ func NewUpdateInstancesError(instanceUpdateErrors []error) *UpdateInstancesError
 func (r *RollingUpgradeReconciler) UpdateInstances(ctx *context.Context,
 	ruObj *upgrademgrv1alpha1.RollingUpgrade,
 	instances []*autoscaling.Instance,
-	currentLaunchConfigName string,
+	launchDefinition *launchDefinition,
 	KubeCtlCall string) error {
 
 	totalNodes := len(instances)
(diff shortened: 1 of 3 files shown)


shrinandj:
Looks good to me. But, please add a "Testing Done" to the PR here (ideally to the original commit message itself) where you add the details of what all tests were done. 

e.g.
- Added unit tests. They pass.
- Verified that upgrade-manager completed successfully for LaunchConfig based upgrade.
- Verified that upgrade-manager completed successfully for LaunchTemplate based upgrade.
- Replaced a launchConfig with launchTemplate. Verified that upgrade completed for such a launchTemplate.

shrinandj:
Great work BTW!! :-)

controllers/rollingupgrade_controller.go
```diff
@@ -489,6 +513,21 @@ func (r *RollingUpgradeReconciler) runRestack(ctx *context.Context, ruObj *upgra
 	return processedInstances, nil
 }
 
+func (r *RollingUpgradeReconciler) resolveLaunchDefinition(asg *autoscaling.Group) (*launchDefinition, error) {
```
eytan-avisror: Does it make more sense to have a single return with both fields?
```
return &launchDefinition{
			kind:                    launchKindLaunchConfiguration,
			launchConfigurationName: asg.LaunchConfigurationName,
                         launchTemplate: asg.LaunchTemplate,
```
If it's Nil, it will stay Nil anyway :) and then you don't have duplicate code.
If you need the context of which one it is downstream, you can have helper predicates `isLaunchConfig()` and `isLaunchTemplate()`

controllers/rollingupgrade_controller.go
```diff
@@ -875,3 +908,35 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
 	ch <- nil
 	return
 }
+
+func requiresRefresh(ec2Instance *autoscaling.Instance, definition *launchDefinition) bool {
+	if definition.kind == launchKindLaunchConfiguration {
+		if ec2Instance.LaunchConfigurationName != nil && *ec2Instance.LaunchConfigurationName != *definition.launchConfigurationName {
```
eytan-avisror: Are we comparing pointers in `*ec2Instance.LaunchConfigurationName != *definition.launchConfigurationName`?
I think this line can be `if aws.StringValue(ec2Instance.LaunchConfigurationName) != nil { ... }`

controllers/rollingupgrade_controller.go
```diff
@@ -875,3 +908,35 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
...
+			return true
+		}
+	} else if definition.kind == launchKindLaunchTemplate {
+		if ec2Instance.LaunchTemplate != nil && definition.launchTemplate != nil {
```
eytan-avisror: Could we simplify this logic? we should think whether all these conditions are really required for us to know a launchtemplate is defined.
Is it not enough to simply say `else if aws.StringValue(ec2Instance.LaunchTemplate) != nil { ... }`? that way we can get rid of all the code below.

controllers/suite_test.go
```diff
@@ -48,7 +48,7 @@ func TestAPIs(t *testing.T) {
 }
 
 var _ = BeforeSuite(func(done Done) {
-	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+	logf.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(GinkgoWriter)))
```
eytan-avisror: This can probably be reverted once you feel this PR is ready to go in

controllers/rollingupgrade_controller.go
uthark: good catch.

controllers/rollingupgrade_controller.go
uthark: LaunchTemplate is an object that have 3 fields - `LaunchTemplateName`, `LaunchTemplateId` and `Version`. So, need to check for the difference. Also, I'm testing what happens if LaunchTemplate refers to version via `$Default` and `$Latest` - additional checks may be required to determine if `$Latest` was updated.

controllers/suite_test.go
uthark: Method `zap.LoggerTo` is deprecated. My PR includes updates, probably, worth to make a separate PR for it.

controllers/rollingupgrade_controller.go
uthark: ASGs can also use [MixedInstancesPolicy](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_MixedInstancesPolicy.html), so, need to make sure code can be easily extended later.

controllers/rollingupgrade_controller.go
eytan-avisror: With LaunchConfigurations, when it drifts, the instances have the field `LaunchConfigurationName` as nil, so we can simply rely on that field being nil.
Is this not the same case with `LaunchTemplate`? when there is a drift, they should strip out that field and you can rely on it in the same manner.

controllers/rollingupgrade_controller.go
eytan-avisror: Can you elaborate on how `MixedInstancesPolicy` might change things? you can still have a single return with all required fields and check if specific ones are nil downstream if it's required.
Also, we might want to store the values so it's easier later e.g. `launchConfigurationName: aws.StringValue(asg.LaunchConfigurationName)`

controllers/suite_test.go
eytan-avisror: Cool, but probably `UseDevMode`should be false?

uthark: @eytan-avisror Made separate PR for that https://github.com/keikoproj/upgrade-manager/pull/44.
Also, old code uses `true`, I just kept same value.

controllers/rollingupgrade_controller.go
uthark: @eytan-avisror I've added support for `MixedInstancesPolicy` too, turned out we use it in some clusters already.

controllers/rollingupgrade_controller.go
uthark: Fixed.

uthark:
@eytan-avisror Could you please take another look?

eytan-avisror:
@uthark Thanks, it looks good. did you get a chance to run this code and do a basic test?

uthark:
Yes, I tested actual rolling update in our dev infra.

shrinandj:
Thanks once again for doing this @uthark !

P.S. I edited the title of the PR and removed the "[WIP]".

PR #44, opened 2020-01-14
Migrate deprecated code.

uthark:
Zap logger deprecated it here: https://github.com/kubernetes-sigs/controller-runtime/pull/560

### DIFF
--- a/main.go
+++ b/main.go
@@ -51,7 +51,7 @@ func main() {
 	flag.IntVar(&maxParallel, "max-parallel", 10, "The max number of parallel rolling upgrades")
 	flag.Parse()
 
-	ctrl.SetLogger(zap.Logger(true))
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -51,7 +51,7 @@ func main() {
 	flag.IntVar(&maxParallel, "max-parallel", 10, "The max number of parallel rolling upgrades")
 	flag.Parse()
 
-	ctrl.SetLogger(zap.Logger(true))
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(os.Stderr)))
```
uthark: We might set it to `false` here, but use `true` in tests.

uthark:
@eytan-avisror Could you please take a look?

PR #46, opened 2020-01-16
Use go 1.13 for building

uthark:
Noticed, that semaphore uses old version of golang.
Bumped to latest 1.13.

### DIFF
--- a/.semaphore/setup.sh
+++ b/.semaphore/setup.sh
@@ -1,13 +1,11 @@
-cache restore $SEMAPHORE_PROJECT_NAME-dep
+cache restore $SEMAPHORE_PROJECT_NAME-dep-$(checksum .semaphore/setup.sh)
 # checks if the packages are already installed
-if [ ! -d '/packages' ]; then
+if [ ! -d '/packages' -o ! -f '/packages/go1.13.6.linux-amd64.tar.gz' ]; then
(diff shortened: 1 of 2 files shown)


.semaphore/setup.sh
```diff
@@ -1,13 +1,11 @@
 cache restore $SEMAPHORE_PROJECT_NAME-dep
 # checks if the packages are already installed
-if [ ! -d '/packages' ]; then
+if [ ! -d '/packages' -a ! -f '/packages/go1.13.6.linux-amd64.tar.gz' ]; then
```
kianjones4: I think this should be `-o` instead of `-a`. This block isn't getting executed and the packages aren't being uploaded since the /packages directory exists and doesn't contain the go1.13.6 package

uthark: Fixed.

uthark: But it still breaks. I see that cache is update, but then the old one is read. Not an expert in SemaphoreCI, do I need to use some extra commands to update cache?

uthark: @kianjones4 Fixed by using checksum of the file as part of the key.

See https://keikoproj-upgrade-manager.semaphoreci.com/branches/e80c4393-75c9-4ec5-83f4-be02d677c071

Issue #45, opened 2020-01-15
Feature Request: Improved rollout strategy


dndungu:
## Challenges
The current rolling upgrade strategy presents the following challenges.
1. Only one instance per AZ can be safely terminated since the strategy is to terminate first, then wait for the ASG to replace terminated instances.
2. Pods can get rescheduled several times during an upgrade. There is no guarantee that pods will get scheduled in the instances using the new Launch Configuration.

## Proposal
1. Create a new ASG with min=0, desired=**n** (some configurable value), max=(same value as old ASG)
1. Wait for the **n** instances to get created and be healthy nodes in the Cluster.
1. Cordon all the instances in the old ASG to prevent the pods from being scheduled in them since we already know they will soon be terminated.
1. Drain **n** instances in the old ASG.
1. Wait for the pods to get successfully scheduled on the new node.
1. For every node successfully drained and until new ASG desired==old ASG desired:
    1. Terminate the drained node.
    1. Increase the desired instances in the new ASG
    1. Wait for the new instance to be created and healthy
    1. Drain another node in the old ASG
    1. Wait for pods to get successfully scheduled in the new pod
1. Delete the old ASG

shrinandj:
@dndungu Thanks for filing this issue. 

The strategy that you mentioned (kinda like blue-green upgrades) had been discussed before. The problem usually is for large ASGs with lots of nodes. In those cases, it may not be possible to spin up **n** new nodes because the AWS account does not have enough instances of that type. Ideally, the code should handle this by simply accepting whatever instances it got and keep the upgrade rolling with those many instances in each "batch". The safest and most reliable approach was to terminate instance 1 by 1 and that's why it was the first one implemented.

There are some more strategies wherein multiple instances are terminated in one go.

The strategy that you propose is certainly efficient and definitely doable. It just needs to handle the corner cases better.

PRs welcome ;-)

eytan-avisror:
In addition, this might be problematic with things like `Kops` which need to be the provisioner of scaling groups, this will be very hard to implement generically, where when we just kill instances we kinda keep it simple.

dndungu:
@shrinandj, thanks for the response, we have faced the AWS capacity issues a number of times. 

A possible solution could be to use create a capacity reservation before starting the rollout and cancel it after a successful rollout. This could also be coupled with having fallback instance types to be used when AWS does not have the desired instance type available.

uthark:
[...]

PR #49, opened 2020-01-18
Eager mode: Launch Before Terminate

eytan-avisror:
Contributes to #45 

This implements an addition to `UpdateStrategy` where you can now add `UpdateStrategy.Mode` to either be `Lazy` (default) or `Eager`.
`Lazy` will use the same approach we currently have of terminating first and launching after.
`Eager` will put an instance in standby to cause scale up and only then proceed with termination sequence.

Manual tests this seems to work fine, still need to add unit tests for this code

TBD:
- [x] Unit Tests

Example Spec:
```
apiVersion: upgrademgr.keikoproj.io/v1alpha1
kind: RollingUpgrade
metadata:
  name: my-rollup
  namespace: kube-system
spec:
  asgName: my-asg.cluster.k8s.local
  nodeIntervalSeconds: 300
  postDrainDelaySeconds: 300
  region: us-west-2
  strategy:
    mode: eager
    drainTimeout: 600
    type: randomUpdate
    maxUnavailable: "100%"
```


controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+			}
+			log.Warnf("%v: cannot set instance %v to stand-by, instance in state %v", ruObj.Name, instanceID, aws.StringValue(instance.LifecycleState))
+			return nil
+		}
+	}
+	_, err = r.ASGClient.EnterStandby(input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			if strings.Contains(aerr.Message(), "not found") {
+				log.Printf("Instance %s not found. Moving on\n", instanceID)
```
shrinandj: Add the `ruObj.Name` prefix to the log message?

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+	if err != nil {
+		return err
+	}
+
+	for _, instance := range r.Asg.Instances {
+		if aws.StringValue(instance.InstanceId) == instanceID {
+			if aws.StringValue(instance.LifecycleState) == autoscaling.LifecycleStateInService {
+				break
+			} else if aws.StringValue(instance.LifecycleState) == autoscaling.LifecycleStateStandby {
+				// already in standby
```
shrinandj: Should we log a message stating that the instance is already in standby?

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+	_, err = r.ASGClient.EnterStandby(input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			if strings.Contains(aerr.Message(), "not found") {
+				log.Printf("Instance %s not found. Moving on\n", instanceID)
+				return nil
+			}
+			switch aerr.Code() {
+			case autoscaling.ErrCodeResourceContentionFault:
+				log.Warn(autoscaling.ErrCodeResourceContentionFault, aerr.Error())
```
shrinandj: Should we add some more info to the message? `ruObj.Name` at least?

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+				return nil
+			default:
+				log.Println(aerr.Error())
```
shrinandj: Let's add some more details to the log message.

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+func (r *RollingUpgradeReconciler) SetStandby(ruObj *upgrademgrv1alpha1.RollingUpgrade, instanceID string) error {
+
+	log.Infof("%v: Setting %v to stand-by", ruObj.Name, instanceID)
+	input := &autoscaling.EnterStandbyInput{
+		AutoScalingGroupName:           aws.String(ruObj.Spec.AsgName),
+		InstanceIds:                    aws.StringSlice([]string{instanceID}),
+		ShouldDecrementDesiredCapacity: aws.Bool(false),
+	}
+
+	err := r.populateAsg(ruObj)
```
shrinandj: Is this call to `populateASG` needed? It will make AWS API calls and might get throttled, etc.

`populateAsg` already stores the result it got from AWS into `ruObj`. Should we just use that?

controllers/rollingupgrade_controller.go
```diff
@@ -294,9 +403,10 @@ func (r *RollingUpgradeReconciler) TerminateNode(ruObj *upgrademgrv1alpha1.Rolli
 		}
 	}
 
-	log.Printf("%s: Termination output: %v\n", ruObj.Name, result)
+	log.Infof("%v: terminated instance %v", ruObj.Name, instanceID)
```
shrinandj: These two lines can be consolidated into a single log message.

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
 	errChan <- nil
 }
 
-func (r *RollingUpgradeReconciler) WaitForTermination(nodeName string, nodeInterface v1.NodeInterface) (bool, error) {
+func (r *RollingUpgradeReconciler) WaitForDesiredInstances(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
+	var (
+		started = time.Now()
+	)
+	for {
+		if time.Since(started) >= (time.Second * time.Duration(ScaleoutTimeoutSeconds)) {
```
shrinandj: Nice!

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
 	errChan <- nil
 }
 
-func (r *RollingUpgradeReconciler) WaitForTermination(nodeName string, nodeInterface v1.NodeInterface) (bool, error) {
+func (r *RollingUpgradeReconciler) WaitForDesiredInstances(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
+	var (
+		started = time.Now()
+	)
+	for {
```
shrinandj: Can we use inverse-exponential-backoff here?

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+		if time.Since(started) >= (time.Second * time.Duration(ScaleoutTimeoutSeconds)) {
+			err := errors.New("WaitForDesired timed out while waiting for instance to be added")
+			return err
+		}
+		err := r.populateAsg(ruObj)
+		if err != nil {
+			return err
+		}
+		if getInServiceCount(r.Asg.Instances) == aws.Int64Value(r.Asg.DesiredCapacity) {
+			break
```
shrinandj: Would be useful to have a log message here stating that the Desired count was reached in the ASG.

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+		time.Sleep(time.Duration(ScaleoutSleepIntervalSeconds) * time.Second)
+	}
+	return nil
+}
+
+func (r *RollingUpgradeReconciler) WaitForDesiredNodes(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
+	var (
+		started = time.Now()
+	)
+	for {
```
shrinandj: Can we use inverse-exponential-backoff here?

controllers/rollingupgrade_controller.go
```diff
@@ -244,28 +248,133 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
...
+}
+
+func (r *RollingUpgradeReconciler) WaitForTermination(ruObj *upgrademgrv1alpha1.RollingUpgrade, nodeName string, nodeInterface v1.NodeInterface) (bool, error) {
 	var (
 		started = time.Now()
 	)
 	for {
 		if time.Since(started) >= (time.Second * time.Duration(TerminationTimeoutSeconds)) {
-			log.Println("WaitForTermination timed out while waiting for node to unjoin")
+			log.Printf("%v: WaitForTermination timed out while waiting for node to unjoin", ruObj.Name)
```
shrinandj: Thanks for fixing this!

controllers/rollingupgrade_controller.go
```diff
@@ -795,6 +910,62 @@ Loop:
...
+
+	// Set instance to standby.
+	err := r.SetStandby(ruObj, targetInstanceID)
+	if err != nil {
+		ch <- err
+		return
+	}
+
+	// Wait for new instance to be created
+	err = r.WaitForDesiredInstances(ruObj)
```
shrinandj: Setting an instance into standby is best-effort. AWS might be at it's absolute limits for the account and may not be able to spin-up new instances. Even if the desired count was not reached, upgrade should proceed. WDYT?

controllers/rollingupgrade_controller.go
eytan-avisror: I was just moving this code around, but will take care of it

controllers/rollingupgrade_controller.go
eytan-avisror: same

controllers/rollingupgrade_controller.go
eytan-avisror: Unfortunately we have to call this for waiters.
Otherwise we will never detect a state has changed.

For example, we are waiting for N instances to be in-service, we must keep refreshing the "cached" version.

In this case (in SetStandby) not sure what was the reason for adding this here.
I will test without it and see if it can be removed

controllers/rollingupgrade_controller.go
eytan-avisror: I don't think we need the termination output, do we? clutters up to logs

controllers/rollingupgrade_controller.go
eytan-avisror: Yes, great idea

controllers/rollingupgrade_controller.go
eytan-avisror: Yes
I think we need additional error handling right after `r.UpdateInstanceEager` or `r.DrainTerminate` are called.

The scenario where you wait for instances but they never show up is not easy to reproduce for testing with very short timeout.

eytan-avisror: your assumption is correct from code perspective, if we time out, we move on and drain/kill, so this makes `eager` mode best-effort like you say.

PR #55, opened 2020-02-20
Check version by value, not by pointers

uthark:
Found out that check if version changed was incorrect and caused rollouts when it should not.

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -979,6 +979,8 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
 		r.Update(*ctx, ruObj)
 		ch <- nil
 		return
+	} else {
+		log.Printf("Will replace instance: %s", targetInstanceID)
 	}
 
 	nodeName := r.getNodeName(i, r.NodeList, ruObj)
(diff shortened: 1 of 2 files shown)


uthark:
@eytan-avisror Could you please take a look?

controllers/rollingupgrade_controller.go
```diff
@@ -1050,7 +1052,7 @@ func requiresRefresh(ec2Instance *autoscaling.Instance, definition *launchDefini
 		if aws.StringValue(instanceLaunchTemplate.LaunchTemplateName) != aws.StringValue(targetLaunchTemplate.LaunchTemplateName) {
 			return true
 		}
-		if aws.String(*instanceLaunchTemplate.Version) != aws.String(*targetLaunchTemplate.Version) {
+		if aws.StringValue(instanceLaunchTemplate.Version) != aws.StringValue(targetLaunchTemplate.Version) {
```
shrinandj: Can there be a unit test for this?

uthark: @shrinandj Added.

uthark:
@eytan-avisror Thank you for approving! Could you please merge it? 
> Only those with write access to this repository can merge pull requests.

PR #60, opened 2020-02-28
DONOTMERGE: Set version to v0.6.1 and update CHANGELOG

davemasselink:
I did mark the latest commit with `v0.6.1` tag in my forked repo, but I'm not sure if the tag will be communicated back to `keikoproj` version of repo as a part of this PR.

If not, I may need to ask someone with core repo permissions to carryout the tagging process.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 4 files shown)


shrinandj:
Can we just do v0.7 instead? It's a simpler versioning scheme.

davemasselink:
@shrinandj, yessir... fixing that up now

PR #65, opened 2020-03-06
Fix shadowing of error variable

### LINKED ISSUE #64: Question: How to detect bad AMI/Configuration and stop the rollout?
Question

**What happened**:
If underlying AMI is bad for whatever reason and new instance doesn't join the cluster, what is the best way to detect the issue and stop the rollout?

**Environment**:
- rolling-upgrade-controller version: 0.6.0
- Kubernetes version : 1.16.7

uthark:
Fixes #64.

**Before:**
```
upgrade-manager-6b6c6b6988-g58dh rolling-upgrade-controller time="2020-03-06T22:16:46Z" level=info msg="rollingupgrade-node-us-west-2a: new node has not yet joined the cluster"
upgrade-manager-6b6c6b6988-g58dh rolling-upgrade-controller time="2020-03-06T22:17:01Z" level=info msg="rollingupgrade-node-us-west-2a: new node has not yet joined the cluster"
upgrade-manager-6b6c6b6988-g58dh rolling-upgrade-controller time="2020-03-06T22:17:01Z" level=info msg="rollingupgrade-node-us-west-2a: Running script echo \"Pods at PreDrain:\"\nkubectl get pods --all-namespaces --field-selector spec.nodeName=${INSTANCE_NAME}\n"
upgrade-manager-6b6c6b6988-g58dh rolling-upgrade-controller time="2020-03-06T22:17:02Z" level=info msg="rollingupgrade-node-us-west-2a: Script finished with output: Pods at PreDrain:\nNAMESPACE     NAME                                      READY
```

**Before with extra logging:**
```
upgrade-manager-98555bcf8-thcfp rolling-upgrade-controller time="2020-03-06T23:04:35Z" level=info msg="rollingupgrade-node-us-west-2b: new node has not yet joined the cluster"
upgrade-manager-98555bcf8-thcfp rolling-upgrade-controller time="2020-03-06T23:04:35Z" level=info msg="Got error: <nil>"
upgrade-manager-98555bcf8-thcfp rolling-upgrade-controller time="2020-03-06T23:04:35Z" level=info msg="Got result error: <nil>"
upgrade-manager-98555bcf8-thcfp rolling-upgrade-controller time="2020-03-06T23:04:35Z" level=info msg="rollingupgrade-node-us-west-2b: Running script echo \"Pods at PreDrain:\"\nkubectl get pods --all-namespaces --field-selector spec.nodeName=${INSTANCE_NAME}\n"
```

**After:**

```
upgrade-manager-ff6c8475f-7tgpz rolling-upgrade-controller time="2020-03-06T23:26:14Z" level=info msg="rollingupgrade-node-us-west-2c: new node has not yet joined the cluster"
upgrade-manager-ff6c8475f-7tgpz rolling-upgrade-controller time="2020-03-06T23:26:14Z" level=info msg="Got error: No more retries left"
upgrade-manager-ff6c8475f-7tgpz rolling-upgrade-controller time="2020-03-06T23:26:14Z" level=info msg="Got result error: rollingupgrade-node-us-west-2c: WaitForDesiredNodes timed out while waiting for nodes to join: No more retries left"
```

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -251,9 +251,9 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
 }
 
 func (r *RollingUpgradeReconciler) WaitForDesiredInstances(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
-
 	var err error
-	for ieb, err := iebackoff.NewIEBackoff(WaiterMaxDelay, WaiterMinDelay, 0.5, WaiterMaxAttempts); err == nil; err = ieb.Next() {
+	var ieb *iebackoff.IEBackoff
+	for ieb, err = iebackoff.NewIEBackoff(WaiterMaxDelay, WaiterMinDelay, 0.5, WaiterMaxAttempts); err == nil; err = ieb.Next() {
 		err = r.populateAsg(ruObj)
 		if err != nil {
 			return err
@@ -276,9 +276,9 @@ func (r *RollingUpgradeReconciler) WaitForDesiredInstances(ruObj *upgrademgrv1al
 }
 
 func (r *RollingUpgradeReconciler) WaitForDesiredNodes(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
-
 	var err error
-	for ieb, err := iebackoff.NewIEBackoff(WaiterMaxDelay, WaiterMinDelay, 0.5, WaiterMaxAttempts); err == nil; err = ieb.Next() {
+	var ieb *iebackoff.IEBackoff
+	for ieb, err = iebackoff.NewIEBackoff(WaiterMaxDelay, WaiterMinDelay, 0.5, WaiterMaxAttempts); err == nil; err = ieb.Next() {
 		r.populateNodeList(ruObj, r.generatedClient.CoreV1().Nodes())
 
 		asg, err := r.GetAutoScalingGroup(ruObj.Name)
@@ -777,7 +777,7 @@ func (r *RollingUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *RollingUpgradeReconciler) setStateTag(ruObj *upgrademgrv1alpha1.RollingUpgrade, instanceID string, state string) error {
-	log.Printf("%v: setting instance %v state to %v", ruObj.Name, instanceID, state)
+	log.Printf("%v: setting instance %q state to %q", ruObj.Name, instanceID, state)
 	err := tagEC2instance(instanceID, EC2StateTagKey, state, r.EC2Client)
 	if err != nil {
 		return err
--- a/controllers/rollingupgrade_controller_test.go
+++ b/controllers/rollingupgrade_controller_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"k8s.io/client-go/kubernetes/scheme"
+
 	"github.com/keikoproj/upgrade-manager/pkg/log"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -78,7 +80,7 @@ func TestErrorStatusMarkJanitor(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
 	instance := &upgrademgrv1alpha1.RollingUpgrade{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"}}
 
-	mgr, err := manager.New(cfg, manager.Options{})
+	mgr, err := buildManager()
 	g.Expect(err).NotTo(gomega.HaveOccurred())
 	c = mgr.GetClient()
 
@@ -781,7 +783,7 @@ func TestFinishExecutionCompleted(t *testing.T) {
 	startTime := time.Now()
 	ruObj.Status.StartTime = startTime.Format(time.RFC3339)
 
-	mgr, err := manager.New(cfg, manager.Options{})
+	mgr, err := buildManager()
 	g.Expect(err).NotTo(gomega.HaveOccurred())
 	c = mgr.GetClient()
 
@@ -809,7 +811,7 @@ func TestFinishExecutionError(t *testing.T) {
 
 	ruObj := &upgrademgrv1alpha1.RollingUpgrade{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"}}
 
-	mgr, err := manager.New(cfg, manager.Options{})
+	mgr, err := buildManager()
 	g.Expect(err).NotTo(gomega.HaveOccurred())
 	c = mgr.GetClient()
 	rcRollingUpgrade := &RollingUpgradeReconciler{
@@ -853,7 +855,7 @@ func TestRunRestackSuccessOneNode(t *testing.T) {
 		Spec:       upgrademgrv1alpha1.RollingUpgradeSpec{AsgName: someAsg},
 	}
 
-	mgr, err := manager.New(cfg, manager.Options{})
+	mgr, err := buildManager()
 	g.Expect(err).NotTo(gomega.HaveOccurred())
 	c = mgr.GetClient()
 
@@ -901,7 +903,7 @@ func TestRunRestackSuccessMultipleNodes(t *testing.T) {
 	ruObj := &upgrademgrv1alpha1.RollingUpgrade{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"},
 		Spec: upgrademgrv1alpha1.RollingUpgradeSpec{AsgName: someAsg}}
 
-	mgr, err := manager.New(cfg, manager.Options{})
+	mgr, err := buildManager()
 	g.Expect(err).NotTo(gomega.HaveOccurred())
 	c = mgr.GetClient()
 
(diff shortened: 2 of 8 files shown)


uthark:
@eytan-avisror Please take a look. Old code always had `err = nil` because `for` loop had its own error. 
As a result `errors.Wrapf` returned null (it returns nil for nil error).

eytan-avisror:
Thanks @uthark 
Did this eventually fail the rollingupgrade CR?

uthark:
Without fix - no, it marked rollout as successful.
With fix - it marked rollout as failed.

eytan-avisror:
Thanks @uthark great find / fix.

For my own understanding, can you explain how this happens? I'm looking at the diff, and while understanding this fixes a bug, functionally it seems equivalent, is this a problem with the scope of `err` variable?.

@shrinandj we might need a better interface for IEB to avoid this in the future

uthark:
The issue was that within for loop declaration := created new inner variable named err. Loop worked as expected, but after loop ended inner variable err remains in the inner scope and outer err was never assigned/kept default nil value.

Check this article for more details:
https://golang.org/doc/effective_go.html#redeclaration

eytan-avisror:
I see, are you referring to `asg, err :=` :

https://github.com/keikoproj/upgrade-manager/blob/767370d04f494cad0a1ceeacedbf2d35809d454f/controllers/rollingupgrade_controller.go#L284-L287

as overriding `err = ieb.Next()` here:
https://github.com/keikoproj/upgrade-manager/blob/767370d04f494cad0a1ceeacedbf2d35809d454f/controllers/rollingupgrade_controller.go#L281

so that the result is that all the attempts expire and we exit the loop instead of throwing an error?

uthark:
@eytan-avisror 
https://github.com/keikoproj/upgrade-manager/blob/master/controllers/rollingupgrade_controller.go#L280-L281

err in line 280 and in line 281 are different one, because := is used.

uthark:
Screen Shot 2020-03-06 at 5 29 13 PM
Screen Shot 2020-03-06 at 5 29 22 PM
Screen Shot 2020-03-06 at 5 29 34 PM

uthark:
> Thanks @uthark great find / fix.
> 
> For my own understanding, can you explain how this happens? I'm looking at the diff, and while understanding this fixes a bug, functionally it seems equivalent, is this a problem with the scope of `err` variable?.
> 
> @shrinandj we might need a better interface for IEB to avoid this in the future / other places
> FYI, this implementation is also mentioned in the docs:
> https://github.com/keikoproj/inverse-exp-backoff

The issue will happen only if you want to use `err` outside of the for loop. Your example in repo is valid.

eytan-avisror:
Oh, and then we wrap a nil err?
https://github.com/keikoproj/upgrade-manager/blob/767370d04f494cad0a1ceeacedbf2d35809d454f/controllers/rollingupgrade_controller.go#L310

which returns as nil

uthark:
Other option would be to declare IEBackoff outside of for loop, like this:

```
func (r *RollingUpgradeReconciler) WaitForDesiredInstances(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
	var err error
	var ieb *iebackoff.IEBackoff
	for ieb, err = iebackoff.NewIEBackoff(WaiterMaxDelay, WaiterMinDelay, 0.5, WaiterMaxAttempts); err == nil; err = ieb.Next() {
		err = r.populateAsg(ruObj)

```

In this case our error would contain error from `ieb.Next` at the end.

uthark:
@eytan-avisror Correct.  https://github.com/keikoproj/upgrade-manager/pull/65#issuecomment-596009833

eytan-avisror:
Yeah, I think that was the confusing part for me, for some reason I assumed `errors.Wrapf()` will always emit an error.

I think what fixes it in your other example is that you use `for ieb, err =` instead `for ieb, err :=` in original code, can we do that instead? I think that is slightly nicer 👍

uthark:
@eytan-avisror Updated PR.

eytan-avisror:
Not sure why unit tests are failing, I think something wrong with latest kubebuilder , it seems we always pull latest here:
https://github.com/keikoproj/upgrade-manager/blob/767370d04f494cad0a1ceeacedbf2d35809d454f/.semaphore/setup.sh#L8

uthark:
Latest  is pulled only if cache is outdated. Otherwise cached version is used.

uthark:
@eytan-avisror I'm looking into test failure.

eytan-avisror:
@uthark I think this is an actual test failure if you follow the trace in the panic.
This error can be ignored (it's just log.Error() call).
```
"failed to update status: no kind is registered for the type v1alpha1.RollingUpgrade in scheme \"pkg/runtime/scheme.go:101\"
```

Looks like this 
```
	g.Expect(err.Error()).To(gomega.HavePrefix("Error updating instances, ErrorCount: 1, Errors: ["))
```
is not returning an error any more (it's returning nil), and this is where the panic is coming from
weird part is that I also see it on master branch and previous unit tests were passing...

uthark:
Hi! Yes, it is erroring before. Also, other tests are also failing.

And error happens even on master (tried locally).

I've fixed error - this is because metrics server is on by default.

```
Unexpected error:
            <*errors.errorString | 0xc000933150>: {
                s: "error listening on :8080: listen tcp :8080: bind: address already in use",
            }
            error listening on :8080: listen tcp :8080: bind: address already in use
        occurred
```

Also, one of the last PRs didn't update test: See https://github.com/keikoproj/upgrade-manager/pull/65/files#diff-5ae70699c0d26d21d55a476a0904dffaR990

Fixing scheme registration now.

eytan-avisror:
@uthark thats great, but that failure (due to port conflict) would cause those tests to get skipped.
There are two tests that are actually failing because the upgrade strategy is not set

eytan-avisror:
Add the following to `ruObj`:
```
Strategy: upgrademgrv1alpha1.UpdateStrategy{
				Mode: "eager",
			},
```

In the following tests:
`TestRunRestackTerminateNodeFail` & `TestRunRestackDrainNodeFail`

It's passing for me now

eytan-avisror:
In fact we should probably do it in all tests (or set a default value in actual code), it means all our tests are not really entering this code:
https://github.com/keikoproj/upgrade-manager/blob/767370d04f494cad0a1ceeacedbf2d35809d454f/controllers/rollingupgrade_controller.go#L1011-L1016

uthark:
@eytan-avisror this method should set default strategy: https://github.com/keikoproj/upgrade-manager/blob/master/controllers/rollingupgrade_controller.go#L827-L857

eytan-avisror:
@uthark 
I think it gets skipped because the tests calls `runRestack()` and not `Reconcile()`

eytan-avisror:
Yeah, it's passing now, once we undo the kubebuilder changes we should be good to go? or if we want to stay hard-coded we should use 2.3.0.

Not sure if you want to fix other tests in this PR (to set a default strategy and make sure they enter the code path), or we can also do it in a different PR.
We can probably add a `getRollingUpgradeObj()` method that returns a default ruObj to all tests.

controllers/rollingupgrade_controller.go
```diff
@@ -193,6 +194,8 @@ func (r *RollingUpgradeReconciler) DrainNode(ruObj *upgrademgrv1alpha1.RollingUp
 	drainTimeout int) error {
 	// Running kubectl drain node.
 	err := r.preDrainHelper(ruObj)
+	spew.Println("err := r.preDrainHelper(ruObj)" , err)
```
eytan-avisror: we can probably also remove this one? I guess a remnant from your debugging 😉

uthark:
I'll update kubebuilder to 2.3.0 explicitly.

Another thing I noticed is the following:
```
go test ./api/... ./controllers/... -coverprofile cover.out
ok  	github.com/keikoproj/upgrade-manager/api/v1alpha1	5.410s	coverage: 1.4% of statements
ok  	github.com/keikoproj/upgrade-manager/controllers	13.989s	coverage: 73.9% of statements
```

But if run in verbose mode there are few failing tests:

```
go test ./api/... ./controllers/... -coverprofile cover.out -v | grep FAIL
--- FAIL: TestRunRestackTerminateNodeFail (0.01s)
--- FAIL: TestUpdateInstancesError (0.01s)
--- FAIL: TestUpdateInstancesPartialError (0.01s)
FAIL
```

Although, command still finishes successfully.

controllers/rollingupgrade_controller.go
uthark: This is cleaned up.

controllers/rollingupgrade_controller_test.go
```diff
@@ -2132,6 +2145,14 @@ func TestWaitForTermination(t *testing.T) {
 	g.Expect(unjoined).To(gomega.BeTrue())
 }
 
+func buildManager() (manager.Manager, error) {
+	err := upgrademgrv1alpha1.AddToScheme(scheme.Scheme)
```
uthark: @eytan-avisror This fixes this error:

```
"failed to update status: no kind is registered for the type v1alpha1.RollingUpgrade in scheme \"pkg/runtime/scheme.go:101\"
```

and after scheme registration there is a new one, because the test framework doesn't setup state properly: 


```
"failed to update status: rollingupgrades.upgrademgr.keikoproj.io \"foo\" not found"
```

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/aws/aws-sdk-go v1.25.0
+	github.com/davecgh/go-spew v1.1.1
```
eytan-avisror: Can just remove this one as well?

uthark: Did `go mod tidy`.

eytan-avisror:
Quite the rabbit hole.
Thanks @uthark 
I will review tomorrow and open an issue to fix unit tests

PR #66, opened 2020-03-07
Use existing go version, simplify maintenance

uthark:
* Use go version from docker image provided by semaphore
* Simplify script to improve maintainability - to upgrade kubebuilder only bump version in `setup.sh` is required.


uthark:

uthark:
@eytan-avisror What do you think about it?

PR #67, opened 2020-03-09
Cleanup unneed fields

uthark:
`status` and `creationTimestamp` are populated on save, don't need to send them.


uthark:

eytan-avisror:
@uthark 
Removing status is not a good idea... there are other fields there that are in use.
Also, proper removal should be from v1alpha1 + `make manifests`.

Regarding creationTimestamp - this is all controlled by `controller-gen` and will be re-added when you run `make manifests`.

I think we can skip this change

uthark:
Ah... So, PR should go to the `controller-gen` then..

PR #78, opened 2020-03-17
Allow to use JSON logging

### LINKED ISSUE #58: Write logs in json
FEATURE REQUEST

It would be nice to have logs in json format.
And also, it would be nice if log message would contain RollingUpgrade object details it referred too (for quickly check logs for a particular RollingUpgrade. ( it is done for some log messages, but not for all of them, at least drain it is not:

```
upgrade-manager-6c6bdbfd5d-zvb7d rolling-upgrade-controller time="2020-02-25T21:00:19Z" level=info msg="Creating a context with timeout - 600"
upgrade-manager-6c6bdbfd5d-zvb7d rolling-upgrade-controller time="2020-02-25T21:00:19Z" level=info msg="Invoking kubectl drain for the node - ip-10-218-50-230.ap-southeast-2.compute.internal"
upgrade-manager-6c6bdbfd5d-zvb7d rolling-upgrade-controller time="2020-02-25T21:00:19Z" level=info msg="rollingupgrade-node-ap-southeast-2c: Running script /usr/local/bin/kubectl drain ip-10-218-50-230.ap-southeast-2.compute.internal --ignore-daemonsets=true --delete-local-data=true --force --grace-period=-1"
```

uthark:
closes #58 
Adds JSON logging as requested in https://github.com/keikoproj/upgrade-manager/issues/58

Better logging will be handled in another PR.

**Testing Done**
Started with new flag and without a flag and confirmed it works as expected.


uthark:
@eytan-avisror Could you or some one from your team take a look?

.gitignore
```diff
@@ -25,4 +25,5 @@ bin
 vendor/
 config/default/manager_image_patch.yaml-e
 .vscode/
-cover.html
\ No newline at end of file
+cover.html
+upgrade-manager
```
eytan-avisror: what is this ignoring?

uthark: cover.html - is the generated code coverage file.
upgrade-manager - binary if you run `go build`.

eytan-avisror: Hmm.. technically we should use makefile `make manager` which places binaries in /bin/manager, which is already ignored.
Does it make sense to use that instead? ignoring everything named `upgrade-manager` might be problematic?

uthark: It is possible to set full path like /cover.html /upgrade-manager.

uthark: Anyway, I can revert this change completely, this is not a goal of this PR.

eytan-avisror: Yep, cover.html can stay (it was already there), but I think we should remove that one (`upgrade-manager`) for now as it overrides the makefile behavior.
Otherwise this PR looks good.

uthark:

main.go
```diff
@@ -74,15 +78,17 @@ func main() {
 	var namespace string
 	var maxParallel int
 	var debugMode bool
+	var logMode string
 	flag.BoolVar(&debugMode, "debug", false, "enable debug logging")
+	flag.StringVar(&logMode, "log-format", "text", "Log mode: supported values: text, json. Default: text.")
```
uthark:
```suggestion
	flag.StringVar(&logMode, "log-format", "text", "Log mode: supported values: text, json.")
```

uthark:
@eytan-avisror After you approve, what's the process to merge?

PR #79, opened 2020-03-17
Improve logging - tag messages with RollingUpgrade name and use structured logging.

uthark:
Covers #58.

This is WIP, to collect feedback.
I'm going to do extra testing and some extra cleanup.

**Testing Done**
Performed rollout and confirmed no issues.

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -127,9 +127,9 @@ func (r *RollingUpgradeReconciler) runScript(script string, background bool, ruO
 
 	out, err := exec.Command(ShellBinary, "-c", script).CombinedOutput()
 	if err != nil {
-		r.error(ruObj, err, "Script finished", "output", out)
+		r.error(ruObj, err, "Script finished", "output", string(out))
 	} else {
-		r.info(ruObj, "Script finished", "output", out)
+		r.info(ruObj, "Script finished", "output", string(out))
 	}
 	return string(out), err
 }
@@ -556,12 +556,12 @@ func (r *RollingUpgradeReconciler) runRestack(ctx *context.Context, ruObj *upgra
 		if len(inProgress) == 0 {
 			// Fetch instances to update from node selector
 			instances = nodeSelector.SelectNodesForRestack(r.ClusterState)
-			r.info(ruObj, "selected instances for rotation: %+v", instances)
+			r.info(ruObj, fmt.Sprintf("selected instances for rotation: %+v", instances))
 		} else {
 			// Prefer in progress instances over new ones
 			instances = inProgress
 			inProgress = []*autoscaling.Instance{}
-			r.info(ruObj, "found in progress instances: %+v", instances)
+			r.info(ruObj, fmt.Sprintf("found in progress instances: %+v", instances))
 		}
 
 		if instances == nil {
@@ -608,7 +608,7 @@ func (r *RollingUpgradeReconciler) finishExecution(finalStatus string, nodesProc
 		r.error(ruObj, err, "failed to update status")
 	}
 	r.admissionMap.Delete(ruObj.Name)
-	r.info(ruObj, "Deleted %s from admission map %p", ruObj.Name, &r.admissionMap)
+	r.info(ruObj, "Deleted from admission map ", "admissionMap", &r.admissionMap)
 	return reconcile.Result{}, nil
 }
 
@@ -626,7 +626,7 @@ func (r *RollingUpgradeReconciler) Process(ctx *context.Context,
 
 	if ruObj.Status.CurrentStatus == StatusComplete ||
 		ruObj.Status.CurrentStatus == StatusError {
-		r.info(ruObj, "No more processing", "Object state", ruObj.Status.CurrentStatus)
+		r.info(ruObj, "No more processing", "currentStatus", ruObj.Status.CurrentStatus)
 
 		if exists := ruObj.ObjectMeta.Annotations[JanitorAnnotation]; exists == "" {
 			r.info(ruObj, "Marking object for deletion")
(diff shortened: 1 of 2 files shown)


uthark:
@eytan-avisror Please take a look and provide your feedback.

controllers/rollingupgrade_controller.go
```diff
@@ -151,35 +151,38 @@ func (r *RollingUpgradeReconciler) preDrainHelper(ruObj *upgrademgrv1alpha1.Roll
 // Operates on any scripts that were provided after the draining of the node.
 // kubeCtlCall is provided as an argument to decouple the method from the actual kubectl call
 func (r *RollingUpgradeReconciler) postDrainHelper(ruObj *upgrademgrv1alpha1.RollingUpgrade, nodeName string, kubeCtlCall string) error {
+	logr := r.logger(ruObj)
```
eytan-avisror: Can you define this globally, so that we don't need this line in every method?

controllers/rollingupgrade_controller.go
```diff
@@ -151,35 +151,38 @@ func (r *RollingUpgradeReconciler) preDrainHelper(ruObj *upgrademgrv1alpha1.Roll
...
 	if ruObj.Spec.PostDrain.Script != "" {
-		_, err := runScript(ruObj.Spec.PostDrain.Script, false, ruObj.Name)
+		_, err := r.runScript(ruObj.Spec.PostDrain.Script, false, ruObj)
 		if err != nil {
-			msg := "Failed to run postDrain script: " + err.Error()
-			log.Printf("%s: %s", ruObj.Name, msg)
+			msg := "Failed to run postDrain script: "
+			logr.Error(err, msg)
```
eytan-avisror: If we are sending this error upstream i.e. `return result`, do we also need to log it here? won't it log twice?

controllers/rollingupgrade_controller.go
```diff
@@ -151,35 +151,38 @@ func (r *RollingUpgradeReconciler) preDrainHelper(ruObj *upgrademgrv1alpha1.Roll
...
+	logr.Info("Waiting for postDrainDelay", "postDrainDelay", ruObj.Spec.PostDrainDelaySeconds)
 	time.Sleep(time.Duration(ruObj.Spec.PostDrainDelaySeconds) * time.Second)
 
 	if ruObj.Spec.PostDrain.PostWaitScript != "" {
-		_, err := runScript(ruObj.Spec.PostDrain.PostWaitScript, false, ruObj.Name)
+		_, err := r.runScript(ruObj.Spec.PostDrain.PostWaitScript, false, ruObj)
 		if err != nil {
 			msg := "Failed to run postDrainWait script: " + err.Error()
-			log.Printf("%s: %s", ruObj.Name, msg)
+			logr.Error(err, msg)
```
eytan-avisror: same question as above, will this log twice?

controllers/rollingupgrade_controller.go
```diff
@@ -237,10 +240,12 @@ func (r *RollingUpgradeReconciler) CallKubectlDrain(ctx context.Context, nodeNam
 
 	// kops behavior implements the same behavior by using these flags when draining nodes
 	// https://github.com/kubernetes/kops/blob/7a629c77431dda02d02aadf00beb0bed87518cbf/pkg/instancegroups/instancegroups.go lines 337-340
-	out, err := runScript(kubeCtlCall+" drain "+nodeName+" --ignore-daemonsets=true --delete-local-data=true --force --grace-period=-1", false, ruObj.Name)
+	script := kubeCtlCall + " drain " + nodeName + " --ignore-daemonsets=true --delete-local-data=true --force --grace-period=-1"
```
eytan-avisror: While were changing this, can you use `fmt.Sprintf` instead?

eytan-avisror:
I think you need to find a way to define logger globally or make it part of the Reconciler without needing to define it per `ruObj` everywhere.
In other places it's worth while doing this worth to clean up places where we log an error and return it (which will log it again)

controllers/rollingupgrade_controller.go
uthark: Need to maintain some context then for the `RollingUpgrade`. Is it fine with you?

controllers/rollingupgrade_controller.go
uthark: Most probably, I just replaced logging, didn't cleanup double logging.

controllers/rollingupgrade_controller.go
uthark: Sure.

uthark:
@eytan-avisror Options to simplify are:
1. Build logger and then propagate it everywhere (don't like it, pollutes API)
2. make helper log methods that will accept `ruObj` and write to logs.
3. Instead of ruObj pass another new struct ReconcileContext that will have both ruObj and logger to use. (might be the best one, also allows further expansion)

Or share your opinion on that.

eytan-avisror:
@uthark 
Yeah, 3. would be good but that would be a bigger refactor - we can probably do this down the road as it will prove useful to have a context object per reconcile (not shared across multiple reconciles such as `RollingUpgradeReconciler`).

Can we do 2. for now, and possibly do 3. down the road? WDYT?

uthark:
I'll update to 2 then.

controllers/rollingupgrade_controller.go
uthark: this error get's propagated to `UpdateInstances` method and then used to build `UpdateInstancesError` which then get logged here as a summary error: https://github.com/keikoproj/upgrade-manager/blob/634036207388b0b5b24e057401a1ca32670bd1fa/controllers/rollingupgrade_controller.go#L670-L674

uthark:
@eytan-avisror I've updated logging, please take a look.

uthark:
Do not merge it yet though, I'm going to fix couple more issues.

uthark:
@eytan-avisror I've fixed couple of issues and tried to do a rollout –  all is good except for scripts outputs – they are rendered in base64.

uthark:
Fixed script output logging too.

uthark:
@eytan-avisror Hello! Do you have any other comments regarding this PR?

PR #82, opened 2020-03-30
add support for conditions

almonteb:
conditions enable the usage of the `kubectl wait` command, allowing for
awaiting the state of a rolling upgrade.

this change additionally adds a Complete condition, which is set on
completion of execution, regardless of success or failure.


shrinandj:
Awesome! This looks really interesting. Can you list some use case where `condition` would be used?

almonteb:
@shrinandj, 

enabling automation to await the completion state (via `kubectl wait`) provides a standard interface for signaling job completion.

since the manager also has a maximum number of concurrent reconciles, there may be value in adding a started condition as well.

shrinandj:
Thanks for your contributions!

PR #92, opened 2020-05-12
Enable forceRefresh of nodes.

### LINKED ISSUE #90: Feature Request - Allow flag to force cycle instance even if launch config hasn't change
Feature Request
**What happened**:

Big Fan of the project, but would like to submit a feature request.

I noticed in previous versions the instances would cycle regardless whether of the underlying launch config has changed or not. In the recent releases, this behavior was fixed to only cycle instances when the launch config has changed. 
In most cases we use upgrade manager to cycle instances every 30 days for new AMI releases and intended behavior works just fine. But in other cases, where we want to cycle nodes in our cluster to pickup non-launch config changes (DNS, agent refreshes, etc) or just cycle the nodes to refresh the environment, it would be nice have a "force-cycle" parameter in the CRD to tell Upgrade Manager to cycle instances regardless if there is a launch config change or not. 
Let me know what you think.

Thanks.

shrinandj:
This commit enables force refreshing of nodes even if the launch-config or launch-template
has not changed.

Testing Done:
- Unit tests
- Ran default configuration with change to launch-config. Verified that the rolling upgrade
  happened and succeeded.
- Ran default config without change to launch-config. Verified that the controller detected
  the identical configs and did not perform node rotation.
- Enabled forceRefresh. Verified that nodes were rotated.

Closes #90

### DIFF
--- a/api/v1alpha1/rollingupgrade_types.go
+++ b/api/v1alpha1/rollingupgrade_types.go
@@ -40,16 +40,18 @@ type PostTerminateSpec struct {
 
 // RollingUpgradeSpec defines the desired state of RollingUpgrade
 type RollingUpgradeSpec struct {
-	PostDrainDelaySeconds int               `json:"postDrainDelaySeconds,omitempty"`
-	NodeIntervalSeconds   int               `json:"nodeIntervalSeconds,omitempty"`
+	PostDrainDelaySeconds int `json:"postDrainDelaySeconds,omitempty"`
+	NodeIntervalSeconds   int `json:"nodeIntervalSeconds,omitempty"`
 	// AsgName is AWS Autoscaling Group name to roll.
-	AsgName               string            `json:"asgName,omitempty"`
-	PreDrain              PreDrainSpec      `json:"preDrain,omitempty"`
-	PostDrain             PostDrainSpec     `json:"postDrain,omitempty"`
-	PostTerminate         PostTerminateSpec `json:"postTerminate,omitempty"`
-	Strategy              UpdateStrategy    `json:"strategy,omitempty"`
+	AsgName       string            `json:"asgName,omitempty"`
+	PreDrain      PreDrainSpec      `json:"preDrain,omitempty"`
+	PostDrain     PostDrainSpec     `json:"postDrain,omitempty"`
+	PostTerminate PostTerminateSpec `json:"postTerminate,omitempty"`
+	Strategy      UpdateStrategy    `json:"strategy,omitempty"`
 	// IgnoreDrainFailures allows ignoring node drain failures and proceed with rolling upgrade.
-	IgnoreDrainFailures    bool              `json:"ignoreDrainFailures,omitempty"`
+	IgnoreDrainFailures bool `json:"ignoreDrainFailures,omitempty"`
+	// ForceRefresh enables draining and terminating the node even if the launch config/template hasn't changed.
+	ForceRefresh bool `json:"forceRefresh,omitempty"`
 }
 
 // RollingUpgradeStatus defines the observed state of RollingUpgrade
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -1066,25 +1066,26 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
 	ch <- nil
 }
 
-func (r *RollingUpgradeReconciler) getNodeCreationTimestamp(ec2Instance *autoscaling.Instance) time.Time {
+func (r *RollingUpgradeReconciler) getNodeCreationTimestamp(ec2Instance *autoscaling.Instance) (bool, time.Time) {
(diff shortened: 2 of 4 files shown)


api/v1alpha1/rollingupgrade_types.go
```diff
@@ -40,16 +40,18 @@ type PostTerminateSpec struct {
...
+	AsgName       string            `json:"asgName,omitempty"`
+	PreDrain      PreDrainSpec      `json:"preDrain,omitempty"`
+	PostDrain     PostDrainSpec     `json:"postDrain,omitempty"`
+	PostTerminate PostTerminateSpec `json:"postTerminate,omitempty"`
+	Strategy      UpdateStrategy    `json:"strategy,omitempty"`
 	// IgnoreDrainFailures allows ignoring node drain failures and proceed with rolling upgrade.
-	IgnoreDrainFailures    bool              `json:"ignoreDrainFailures,omitempty"`
+	IgnoreDrainFailures bool `json:"ignoreDrainFailures,omitempty"`
+	// ForceRefresh enables draining and terminating the node even if the launch config/template hasn't changed.
+	ForceRefresh bool `json:"forceRefresh,omitempty"`
```
eytan-avisror, line 54: Does it make more sense to do this as an annotation? 
That way you can completely remove the annotation when you are done, and there are no CRD/API changes required?

controllers/rollingupgrade_controller.go
```diff
@@ -1000,8 +1001,7 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
 	// If the running node has the same launchconfig as the asg,
 	// there is no need to refresh it.
 	targetInstanceID := aws.StringValue(i.InstanceId)
-	if !requiresRefresh(i, launchDefinition) {
-		r.info(ruObj, "Ignoring since it has the correct launch-config", "instanceID", targetInstanceID)
+	if !r.requiresRefresh(ruObj, i, launchDefinition) {
```
eytan-avisror, line 1004: Won't this mean if the controller gets restarted and reloads the specs it will refresh again?
we should set it to false once we are done, or use an annotation and just delete it.

shrinandj: Great point ..

api/v1alpha1/rollingupgrade_types.go
shrinandj: We want this to be part of the main functionality and having this as a first class option in the CR makes is that way. IMHO, annotations are best suited for auxiliary or implementation specific options (e.g. aws specific options in an Ingress object). `forceRefresh` is not such an option.

controllers/rollingupgrade_controller.go
shrinandj: One option is to annotate all the existing nodes if the `forceRefresh` flag is set and then look for whether the annotation exists when the node is being processed. This has a couple of drawbacks. (i) If there are large number of nodes in the ASG, all these will require an update thereby adding that much more work to do on the APIServer and (ii) the implementation will be fragmented between processing the `forceRefresh` flag initially and annotating all the nodes and then checking for each one of the nodes.

How about this:

- Every object in Kubernetes has a `creationTimestamp`. The node object has a creationTimestamp and the CR itself also has a creationTimestamp.
- Within `refreshRequired`, if the `forceRefresh` flag is set, there can be an additional check that the `creationTimestamp` of the node should be less than the `creationTimeStamp` of the CR. If the node was created after the CR was created, it won't be part of the refresh.

WDYT?

eytan-avisror: 👍 I think that should work.

controllers/rollingupgrade_controller.go
```diff
@@ -1066,24 +1066,53 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
...
+	}
+	return time.Time{}
+}
+
+func (r *RollingUpgradeReconciler) requiresRefresh(ruObj *upgrademgrv1alpha1.RollingUpgrade, ec2Instance *autoscaling.Instance,
+	definition *launchDefinition) bool {
+
+	if ruObj.Spec.ForceRefresh {
+		nodeCreationTS := r.getNodeCreationTimestamp(ec2Instance)
+		if nodeCreationTS.Before(ruObj.CreationTimestamp.Time) {
```
eytan-avisror, line 1085: should you check if nodeCreationTS is not zero? what happens if in `getNodeCreationTimestamp` node is not found and it returns `time.Time{}`? it will evaluate as `0001-01-01 00:00:00 +0000 UTC` and that would be considered before.
Maybe you can refactor `getNodeCreationTimestamp` to return (time.Time, bool) and you can do the if condition using that bool, if the node is not found return `false, time.Time{}`:
```
if ts, ok := r.getNodeCreationTimestamp(ec2Instance); ok {
    < .. logic .. >
}
```

shrinandj: Done

PR #94, opened 2020-06-04
add retry for TerminateInstanceInAutoScalingGroup

### LINKED ISSUE #77: Terminate may fail and proceed in case transition to standby takes too long
It was noticed that when the transition to standby mode takes to long, for example if the ASG is configured to deregister a targetgroup / load balancer, the terminate request may fail with `ScalingActivityInProgress` error.
In addition it seems the upgrade proceeds even if the terminate fails

The issues should be fixed:

- In eager mode, we should wait until standby transition is completed to avoid this issue.
- When the termination fails the rollup should fail and halt the upgrade

formuzi:
fix #77 
This PR implement retries when TerminateInstanceInAutoScalingGroup failed

Test done
1. Unit tests
2. Create a cluster, update `nodes` instance-group, create CR and trigger a rolling upgrade. Verify retry work as expected

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -18,6 +18,8 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"os"
 	"os/exec"
 	"strconv"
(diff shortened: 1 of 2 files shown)


shrinandj:
- Were you able to reproduce the problem where the retry was needed?
- Can we add some unit tests that exercise the retry logic?

PR #96, opened 2020-06-05
Check for CR before updating each instance.

shrinandj:
This takes care of the case when a CR is deleted in the middle of doing rolling-upgrade. If the CR is deleted, no new instances will get drained and terminated. The ones in the middle of drain and terminate will be completed.


shrinandj:
Closing this in favor of https://github.com/keikoproj/upgrade-manager/pull/97.

PR #97, opened 2020-06-09
Check for CR before updating each instance.

### LINKED ISSUE #95: Rolling upgrade processing should stop if CR is deleted midway
FEATURE REQUEST

**What happened**:
- A rolling-upgrade CR was created.
- The controller started processing the CR and nodes started getting drained and terminated as per the strategy
- The CR was deleted by the user with the expectation that the drain and termination of the nodes will stop
- However, this does not happen. The controller threw the following error yet continued to drain and terminate instances:
```
2020-06-05T00:49:18.343Z	ERROR	controllers.RollingUpgrade	failed to update status	{"rollingupgrade": "rollup-nodes-20200605004225", "error": "Operation cannot be fulfilled on rollingupgrades.upgrademgr.keikoproj.io \"rollup-nodes-20200605004225\": StorageError: invalid object, Code: 4, Key: /registry/upgrademgr.keikoproj.io/rollingupgrades/instance-manager/rollup-nodes-20200605004225, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: bbbb09c5-daf7-40e6-8745-e1c2a47dd924, UID in object meta: "}
```

Draining and termination the instances in the CR should stop if the CR is deleted.

**How to reproduce it (as minimally and precisely as possible)**:
See steps above.

**Anything else we need to know?**:

shrinandj:
Commits:

---
  This commit introduces the inProcessASGs map for keeping track of which ASGs are being processed. If an ASG is being processed and a new CR is submitted for the same ASG, the new CR get requeued until the first one is completed.

    This takes care of the case when a CR is deleted in the middle of doing rolling-upgrade. If the CR is deleted, no new instances will get drained and terminated. The ones in the middle of drain and terminate will be completed.

    Testing Done:

    - Adjusted unit tests for the new code. Ensured that unit tests complete successfully.
    - Deleted rollup CR when it was in the middle of being processed. Ensured that the CR completed the node that was being processed. Others nodes were ignored correctly.
    - Deleted rollup CR after it was completed. Ensured that it was deleted.
---

    Ignore "Invalid object" errors during status update.

    These errors are thrown if the CR is deleted while it was being processed. Unfortunately, there is no way other than error string comparison for doing this.

    Testing Done:

    - Unit tests succeed.
    - Started an upgrade. Deleted the CR while being processed. Verified that the correct log messages were logged and the controller didn't crash.

---

    Check admissionMap for every UpdateInstance

    This commit adds a check for an entry in the admissionMap during every UpdateInstance run. If the CR is deleted, the entry for it is removed from the admissionMap. And any calls to UpdateInstance from then on get short-circuited.

    Testing Done:
    - Unit tests passed
    - Basic rolling upgrade worked as expected.

---

    Move the code for checking CR deletion into Reconcile

    This commit moves the code for handling CR deletion from Process() to Reconcile(). If a CR is deleted while it is being processed, the Reconcile() method will be called and the CR is removed from the admissionMap. This will enable the asynchronously running Process() to check whether the CR is deleted periodically.

    Testing Done:

    - Unit tests pass.
    - Created a CR. While it was processing, delete it. Ensured that the code for handling the delete succeeded.

---

    Make Process run asynchronously

    This enables the Reconcile() to be called again. If and when the CR gets modified/deleted, the Reconcile() can now be called and appropriate actions can be taken.

    Testing Done:

    - Unit tests pass
    - Tested basic rolling-upgrade.

---

    Remove return values from "Process()"

    This is in preparation for calling Process() in a go-routine.

    Testing Done:
    - Unit tests pass.

---

    Add rolling upgrade CR name to admissionMap before processing starts

    This ensures that the admissionMap could be used in subsequent go-routines. In later commits, the admissionMap will be updated to indicate that the CR has been deleted. The go-routines that update the instances will check this and choose to proceed or not.

    Testing Done:

    - Unit tests pass. Ran a simple upgrade. It succeeded.

---

    Mark an instance as "completed" iff it was "in-progress"

    If an instance was started to get updated, it was marked "in-progress". However, there are cases when the instance is never marked "completed". This commit fixes that.

    All instance are marked "completed" if they started their upgrade.

    Testing Done:

    - Unit tests for this exist. Modified them to confirm expected behavior.
Closes #95

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -1000,13 +1000,22 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
 	launchDefinition *launchDefinition,
 	KubeCtlCall string,
 	ch chan error) {
+	targetInstanceID := aws.StringValue(i.InstanceId)
 	// If an instance was marked as "in-progress" in ClusterState, it has to be marked
 	// completed so that it can get considered again in a subsequent rollup CR.
-	defer r.ClusterState.markUpdateCompleted(*i.InstanceId)
+	defer r.ClusterState.markUpdateCompleted(targetInstanceID)
+
+	// Check if the rollingupgrade object still exists
+	_, ok := r.admissionMap.Load(ruObj.Name)
+	if !ok {
+		r.info(ruObj, "Object either force completed or deleted. Ignoring node update")
+		ruObj.Status.NodesProcessed = ruObj.Status.NodesProcessed + 1
+		ch <- nil
+		return
+	}
 
 	// If the running node has the same launchconfig as the asg,
 	// there is no need to refresh it.
-	targetInstanceID := aws.StringValue(i.InstanceId)
 	if !r.requiresRefresh(ruObj, i, launchDefinition) {
 		ruObj.Status.NodesProcessed = ruObj.Status.NodesProcessed + 1
 		if err := r.Status().Update(*ctx, ruObj); err != nil {
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -99,6 +99,7 @@ type RollingUpgradeReconciler struct {
 	ASGClient       autoscalingiface.AutoScalingAPI
 	generatedClient *kubernetes.Clientset
 	NodeList        *corev1.NodeList
+	inProcessASGs   sync.Map
 	admissionMap    sync.Map
 	ruObjNameToASG  sync.Map
 	ClusterState    ClusterState
@@ -544,7 +545,7 @@ func (r *RollingUpgradeReconciler) runRestack(ctx *context.Context, ruObj *upgra
 
 	nodeSelector := getNodeSelector(asg, ruObj)
 
-	// set the state of instances in the ASG to new in the cluster store
+	r.inProcessASGs.Store(*asg.AutoScalingGroupName, "running")
 	r.ClusterState.initializeAsg(*asg.AutoScalingGroupName, asg.Instances)
 	defer r.ClusterState.deleteEntryOfAsg(*asg.AutoScalingGroupName)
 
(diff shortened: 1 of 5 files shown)


controllers/rollingupgrade_controller.go
```diff
@@ -701,10 +701,11 @@ func (r *RollingUpgradeReconciler) Process(ctx *context.Context,
 	nodesProcessed, err := r.runRestack(ctx, ruObj, KubeCtlBinary)
 	if err != nil {
 		r.error(ruObj, err, "Failed to runRestack")
-		return r.finishExecution(StatusError, nodesProcessed, ctx, ruObj)
+		r.finishExecution(StatusError, nodesProcessed, ctx, ruObj)
+		return
 	}
 
-	return r.finishExecution(StatusComplete, nodesProcessed, ctx, ruObj)
+	r.finishExecution(StatusComplete, nodesProcessed, ctx, ruObj)
```
eytan-avisror, line 708: Does it make more sense to defer this call

controllers/rollingupgrade_controller.go
```diff
@@ -1054,20 +1077,16 @@ func (r *RollingUpgradeReconciler) UpdateInstance(ctx *context.Context,
 		r.info(ruObj, "termination waiter completed but node is still joined, will proceed with upgrade", "nodeName", nodeName)
 	}
 
-	// set the EC2 tag indicating the state to completed
-	err = r.setStateTag(ruObj, targetInstanceID, "completed")
```
eytan-avisror, line 1058: Why are we removing this?

shrinandj: The instance is already terminated. So this call always fails

controllers/rollingupgrade_controller.go
shrinandj: Good idea. Done.

shrinandj: Actually.. maybe I spoke too soon. There is at least one case where we don't want to call finishExecution; when the CR is already in completed state. To get past this, we could do `defer finishExecution` after the check. But then there are other parameters needed for finishExecution that need to be set correctly: status, nodesProcessd. This makes the code complex and hard to read. One has to remember to set these variables to the correct value before the function ends.

Let's keep this as is for now. There is a need to refactor a bunch of this code separately.

controllers/rollingupgrade_controller.go
eytan-avisror: You can still tag a terminated instance, this is required for the resume feature

PR #112, opened 2020-08-19
Validation step to check Nodes and ASG launch configs

### LINKED ISSUE #108: Add validation step at the end of rolling-upgrade
FEATURE REQUEST

At the end of the rolling-upgrade before marking the entire CR complete, it would be good idea to run a simple validation step. This validation step should list all the nodes from the ASG and ensure that they are using the correct launch-config/launch-template.

hard-fault:
Fixes #108 

Validation pass scenario:
pass

Validation fail scenario:
fail


controllers/rollingupgrade_controller.go
```diff
@@ -704,9 +705,43 @@ func (r *RollingUpgradeReconciler) Process(ctx *context.Context,
...
+	//Get ASG launch config
+	err := r.populateAsg(ruObj)
+	if err != nil {
+		return errors.New("Unable to populate the ASG object")
+	}
+	asg, err := r.GetAutoScalingGroup(ruObj.Name)
+	if err != nil {
+		return fmt.Errorf("Unable to load ASG with name: %s", ruObj.Name)
+	}
+	launchDefinition := NewLaunchDefinition(asg)
```
shrinandj, line 731: LaunchDefinition can have a `launchConfig` or a `launchTemplate` inside it. We'll have to handle both the cases here.

controllers/rollingupgrade_controller.go
```diff
@@ -704,9 +705,43 @@ func (r *RollingUpgradeReconciler) Process(ctx *context.Context,
...
+	launchConfigASG := *(launchDefinition.launchConfigurationName)
+
+	//Get ec2 instances and their launch configs.
+	ec2instances := asg.Instances
+	for _, ec2Instance := range ec2instances {
+		if launchConfigASG != aws.StringValue(ec2Instance.LaunchConfigurationName) {
```
shrinandj, line 737: Can you check of the `ec2Instance` object has a `LaunchTemplateID` in it?

Basically, if the launchDefinition has a `launchConfig`, the above check is correct. If the launchDefinition has a `launchTemplate`, the check should be modified for that.

shrinandj:
Looks good for the handling of launch configurations. Can you add some unit tests as well?

controllers/rollingupgrade_controller.go
```diff
@@ -444,6 +444,7 @@ func (r *RollingUpgradeReconciler) getNodeName(i *autoscaling.Instance, nodeList
 	return node.Name
 }
 
+//TODO(sbadiger): Why are we listing all the nodes here and not just the nodes from ASG? Use asg.Instances instead of nodeList
```
shrinandj, line 447: Looks like a leftover comment from design/dev phase. Let's nuke it.

controllers/rollingupgrade_controller.go
hard-fault: ec2Instance doesn't have Launch template when accessed through **asg.Instances**
`"Instance": "{AvailabilityZone: "xx-yy", HealthStatus: "Healthy", InstanceId: "i-abc", InstanceType: "aa.bb",  LaunchConfigurationName: "nodes-xxyy",  LifecycleState: "InService",  ProtectedFromScaleIn: false}"`

However, when I access the ec2Instance using **NodeSelector()**, I see both launch config and launch template fields.
`"Instance": [{"AvailabilityZone":"us-west-2c","HealthStatus":"Healthy","InstanceId":"i-xxx","InstanceType":"xx.yy","LaunchConfigurationName":"sbadiger-instance-manager-nodes-xxx","LaunchTemplate":null,"LifecycleState":"InService","ProtectedFromScaleIn":false,"WeightedCapacity":null}]`

controllers/rollingupgrade_controller.go
hard-fault: Yes and No. **nodeList** has all the nodes out of ASG which will always fail the if condition inside the function. Maybe we can track this through another issue.

controllers/rollingupgrade_controller.go
```diff
@@ -704,9 +704,50 @@ func (r *RollingUpgradeReconciler) Process(ctx *context.Context,
...
+	launchDefinition := NewLaunchDefinition(asg)
+	launchConfigASG, launchTemplateASG := launchDefinition.launchConfigurationName, launchDefinition.launchTemplate
+
+	//Get ec2 instances and their launch configs.
+	ec2instances := asg.Instances
+	for _, ec2Instance := range ec2instances {
+		ec2InstanceID, ec2InstanceLaunchConfig, ec2InstanceLaunchTemplate := ec2Instance.InstanceId, ec2Instance.LaunchConfigurationName, ec2Instance.LaunchTemplate
+		if aws.StringValue(launchConfigASG) != aws.StringValue(ec2InstanceLaunchConfig) {
+			r.info(ruObj, "Launch config doesn't match with ASG", "Instance ID", aws.StringValue(ec2InstanceID), "Instance LaunchConfig", aws.StringValue(ec2InstanceLaunchConfig))
+			err = errors.New("Launch config mismatch")
```
eytan-avisror, line 740: There is no need to log an error (line 739) if you are returning an error (it will be logged downstream).
Also, why not just return the error at this point?
In addition, can we use a more descriptive message here?
e.g. "post-validation failed: instances found with mismatching launch configuration" 

Same for launch template below

controllers/rollingupgrade_controller.go
hard-fault: Addressed this issue with latest commit.

controllers/rollingupgrade_controller.go
hard-fault: I thought it might be useful to know which all instances have a mismatch. Hence returning at the end. True. The logging here was also for the same reason(I included the instance ID int the logs)

controllers/rollingupgrade_controller.go
hard-fault: Addressed this in the latest commit. LaunchTemplateID is also validated.

controllers/rollingupgrade_controller.go
hard-fault: Removed the comment.

controllers/rollingupgrade_controller.go
kevdowney: If you're relying on upstream logging, the error would be logged as message so you could be detailed like so...
```suggestion
			err = fmt.Errorf("launch config mismatch, instance config %s does not match asg config %s", aws.StringValue(ec2InstanceID), aws.StringValue(ec2InstanceLaunchConfig))
```
Also, go errors are not supposed to have capitalization. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Another issue is your in a loop and return here is outside the loop, so it's correct that `return err` should be done within the if blocks or you should collect all errors and return a concatenated list or the complete error slice `[]err`

hard-fault: Ah! I see. I shall incorporate these. Thanks, Kevin!

PR #113, opened 2020-08-24
Adding Functional Test

### LINKED ISSUE #70: Functional Test
Upgrade manager should have a basic functional test that runs through basic code paths.
https://github.com/cucumber/godog might be a good option for scaffolding.

Tests can do a rolling upgrade of eager mode (has more coverage) and should verify things such as:

- If instances fail to join, upgrade should error out
- if instances fail to be added to scaling group, upgrade should error out
- etc

garomonegro:
Closes #70 

Adding:
- Implementation of the BDD
- Github Actions workflow for the BDD
- Badge

Modifying:
- CI workflow to push `master` docker image on "push" events

Notes:
In order to activate the BDD workflow, we might have to do a push to the repo after this PR is merged. Apparently Github Actions workflows on schedules activate after the next push to the repo: https://github.community/t/schedule-workflows-missing/17653


garomonegro:
The `Require status checks to pass before merging` rule under `Branch protection rule` is attached to the name of the job instead of the name of the workflow. I changed the name of the job for the CI workflow and that affects the rule, I will sent a commit with a generic name, e.g. CI, to avoid having to modify those rules in the future. We will have to modify the rule on last time after that change

PR #118, opened 2020-09-01
Set timeout for drain command

shaoxt:
Pass the timeout to drain command so that it will get timeout when it waits on PDB.


eytan-avisror:
@shaoxt 
DrainTimeout is already handled here:
https://github.com/keikoproj/upgrade-manager/blob/master/controllers/rollingupgrade_controller.go#L208-L215

shaoxt:
Right now it has timeout implementation in the process.

PR #134, opened 2020-11-05
During upgrade, ignore terminated instance.

### LINKED ISSUE #133: Error caused by a terminated instance during Upgrade
Bug
**What happened**:
While the upgrade is in progress, if an instance is terminated it still appears in the asg.Instances list. Because of this, one of the API throws error. 

We must not consider this as an error (as the instance is already terminated and doesn't require any further action) and proceed towards the next set of nodes.

```
func getGroupInstanceState(group *autoscaling.Group, id string) (string, error) {
	for _, instance := range group.Instances {
		if aws.StringValue(instance.InstanceId) != id {
			continue
		}
		return aws.StringValue(instance.LifecycleState), nil
	}
	return "", errors.Errorf("could not get instance group state, instance %v not found", id)
}
```

**How to reproduce it (as minimally and precisely as possible)**:
During an upgrade, terminate any instance that hasn't yet been picked by the controller.

shreyas-badiger:
Fixes #133 

- [x] Unit Testing


controllers/rollingupgrade_controller.go
```diff
@@ -375,7 +375,8 @@ func (r *RollingUpgradeReconciler) SetStandby(ruObj *upgrademgrv1alpha1.RollingU
 
 	instanceState, err := getGroupInstanceState(asg, instanceID)
 	if err != nil {
-		return err
+		r.info(ruObj, "Cannot get instance state ", "instanceID", instanceID)
```
kevdowney, line 378: What is the `error` message you get when this happens?

vgunapati, line 378: maybe we should `warn` instead of `info` WDYT?

shrinandj, line 378: Also, will the only reason that `getGroupInstanceState` returns an error be that the instance is gone? We might want to do a more specific check, no?

shreyas-badiger: Yes. The only instance when `getGroupInstanceState` errors out is with a NotFound.
```
func getGroupInstanceState(group *autoscaling.Group, id string) (string, error) {
    for _, instance := range group.Instances {
        if aws.StringValue(instance.InstanceId) != id {
            continue
        }
        return aws.StringValue(instance.LifecycleState), nil
    }
    return "", errors.Errorf("could not get instance group state, instance %v not found", id)
}
```

kevdowney: In this case, I would put the message as `Warning` and log the original error message which already says that instance could not get the state. The implementation shouldn't swallow the error, just push it as warning.

shrinandj: +1 to what @kevdowney said

controllers/rollingupgrade_controller.go
```diff
@@ -375,7 +375,8 @@ func (r *RollingUpgradeReconciler) SetStandby(ruObj *upgrademgrv1alpha1.RollingU
 
 	instanceState, err := getGroupInstanceState(asg, instanceID)
 	if err != nil {
-		return err
+		r.info(ruObj, "WARNING:", "message", err)
```
kevdowney, line 378:
```suggestion
		r.info(ruObj, fmt.sprintf("WARNING: %v", err))
```

kevdowney (approved):

controllers/rollingupgrade_controller.go
```diff
@@ -375,7 +375,8 @@ func (r *RollingUpgradeReconciler) SetStandby(ruObj *upgrademgrv1alpha1.RollingU
 
 	instanceState, err := getGroupInstanceState(asg, instanceID)
 	if err != nil {
-		return err
+		r.info(ruObj, fmt.Sprintf("WARNING: %v", err))
```
shrinandj, line 378: Please file a bug for creating the `warn()` method in `RollingUpgradeReconciler`. This mechanism of using an info level message with `WARNING` as the text in it is not ideal.

shrinandj (approved):

PR #138, opened 2020-11-12
If instance is in standby mode already, just return

uthark:
Saw several times repetitive message about unable to set instance state to standby because it is already in standby.

Other changes:
*  move input creation closer to the usage.
* rename function to improve readability.

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -299,26 +299,31 @@ func (r *RollingUpgradeReconciler) GetAutoScalingGroup(rollupName string) (*auto
 // SetStandby sets the autoscaling instance to standby mode.
 func (r *RollingUpgradeReconciler) SetStandby(ruObj *upgrademgrv1alpha1.RollingUpgrade, instanceID string) error {
 	r.info(ruObj, "Setting to stand-by", ruObj.Name, instanceID)
-	input := &autoscaling.EnterStandbyInput{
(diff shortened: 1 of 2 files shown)


uthark:
@eytan-avisror @shrinandj  Please take a look.

controllers/rollingupgrade_controller.go
```diff
@@ -299,28 +299,33 @@ func (r *RollingUpgradeReconciler) GetAutoScalingGroup(rollupName string) (*auto
...
 	}
 
-	instanceState, err := getGroupInstanceState(asg, instanceID)
+	instanceState, err := getInstanceStateInASG(asg, instanceID)
 	if err != nil {
 		r.info(ruObj, fmt.Sprintf("WARNING: %v", err))
 		return nil
 	}
 
+	if instanceState == autoscaling.LifecycleStateStandby {
```
eytan-avisror, line 314: Isnt this covered by the next block `if !isInServiceLifecycleState(instanceState)` ?
Do you just want to avoid the log entry?

uthark: Yes, I want to avoid log entry. If instance is already in standby, we should not log anything. Also, investigating issue when upgrade doesn't finish.

uthark: Are you ok to remove logging completely then?

shrinandj: @eytan-avisror Should the log message be removed completely? Would it affect debuggability?

shrinandj (approved):

PR #132, opened 2020-11-05
Extract script runner to a separate type; fix work with env. variables

uthark:
Fixes https://github.com/keikoproj/upgrade-manager/issues/129

1. Extract work with scripts to a separate type.
2. Fix work with env. variables.
3. Simplify API by not requiring kubectl binary.

PR #144, opened 2020-11-16
Propagate parent env variables to allow to talk with API Server

uthark:
Discovered that new script runner doesn't propagate env variables from parent process. It prevents to talk to API Server from scripts. This PR fixes it. Related to #132


uthark:
@eytan-avisror and/or @shrinandj please take a look.

shrinandj (approved):

uthark:
And small follow up https://github.com/keikoproj/upgrade-manager/pull/145

we always need to use parent env.

PR #148, opened 2020-11-18
Move cluster state from package level to a cluster state impl

uthark:
Found out that cluster state was public at package level, move it to a ClusterState and make it private.
Other changes: improve code readability.


uthark:
@keikoproj/upgrade-manager-approvers Please take a look.

shrinandj (approved):

PR #151, opened 2020-11-20
Don't uncordon node on failure to run postDrain script when IgnoreDrainFailures set

adammw:
Uncordoning the node when IgnoreDrainFailures is set will result in the node being uncordoned then immediately terminated since the drain failure is then ignored and termination continues anyway. Therefore it would be best not to uncordon it in this situation when the flag is set.

### DIFF
--- a/controllers/rollingupgrade_controller_test.go
+++ b/controllers/rollingupgrade_controller_test.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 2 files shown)


shrinandj (approved):

shrinandj:
You might need to signoff the commit before we can merge this. https://github.com/keikoproj/upgrade-manager/pull/151/checks?check_run_id=1433524811

adammw:
@shrinandj I'm checking with our legal department around the CLA and DCO - https://github.com/keikoproj/keiko#news mentions the project moving from CLA to DCO, is the CLA still required?

shrinandj:
Can a unit test be added for this so that it does not regress later?

adammw:
Added a test by observing the side effect of the command being run. It's a little ugly but it should work.

adammw:
cc @shrinandj

PR #163, opened 2020-12-11
BDD manual trigger 

garomonegro:
Added [`workflow_dispatch`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_dispatch) as an event to trigger the BDD workflow.


uthark:
@garomonegro please rebase your branch on top of `master` to cleanup commits in your PR.

garomonegro:
My bad. Cleaning the PR would be more work than what the PR is actually adding. I will just close it and whoever needs to manually trigger the BDD can add this change later. Thanks for the input!

PR #166, opened 2020-12-17
Expose template list and other execution errors to logs

adammw:
Upon deploying v0.17 to our environment, rollingupgrades were silently failing with no decernable error messages. From tracing the possible error points, we deduced it was due to new launchtemplate requests that the serviceaccount did not have permission for. 

This PR adds an explicit error message when launchtemplate fetching fails similar to what happens in the other populate methods, as well as changing the signature of the private finishExecution method to accept an error instead of a status, allowing the error itself to be logged along with the status.

Open question if the error should also be exposed to the kubernetes event as well, or just remain in the controller logs.

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -455,14 +455,15 @@ func (r *RollingUpgradeReconciler) populateAsg(ruObj *upgrademgrv1alpha1.Rolling
 	return nil
 }
 
-func (r *RollingUpgradeReconciler) populateLaunchTemplates() error {
+func (r *RollingUpgradeReconciler) populateLaunchTemplates(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
 	launchTemplates := []*ec2.LaunchTemplate{}
 	err := r.EC2Client.DescribeLaunchTemplatesPages(&ec2.DescribeLaunchTemplatesInput{}, func(page *ec2.DescribeLaunchTemplatesOutput, lastPage bool) bool {
 		launchTemplates = append(launchTemplates, page.LaunchTemplates...)
 		return page.NextToken != nil
 	})
 	if err != nil {
-		return err
+		r.error(ruObj, err, "Failed to populate launch template list")
+		return fmt.Errorf("failed to populate launch template list for %s: %w", ruObj.NamespacedName(), err)
 	}
 	r.LaunchTemplates = launchTemplates
 	return nil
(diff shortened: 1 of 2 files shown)


controllers/rollingupgrade_controller.go
```diff
@@ -463,14 +463,15 @@ func (r *RollingUpgradeReconciler) populateAsg(ruObj *upgrademgrv1alpha1.Rolling
...
+func (r *RollingUpgradeReconciler) populateLaunchTemplates(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
 	launchTemplates := []*ec2.LaunchTemplate{}
 	err := r.EC2Client.DescribeLaunchTemplatesPages(&ec2.DescribeLaunchTemplatesInput{}, func(page *ec2.DescribeLaunchTemplatesOutput, lastPage bool) bool {
 		launchTemplates = append(launchTemplates, page.LaunchTemplates...)
 		return page.NextToken != nil
 	})
 	if err != nil {
-		return err
+		r.error(ruObj, err, "Failed to populate launch template list")
+		return errors.Wrap(err, ruObj.Name+": Failed to populate launch template list")
```
uthark, line 474: can use `fmt.Errorf`
```suggestion
		return fmt.Errorf("failed to populate launch template list for %q: %w", ruObj.Name, err)
```

uthark (approved):

controllers/rollingupgrade_controller.go
shrinandj: @adammw , let's commit this suggestion so that we can go ahead and merge this change

adammw: any reason to use that over the error logger? it appears to give a better result as it'll automatically add the stacktrace and other niceities

adammw: sorry, thought this was a comment on the logger line not error wrapping. updated.

controllers/rollingupgrade_controller_test.go
```diff
@@ -970,7 +972,8 @@ func TestFinishExecutionError(t *testing.T) {
 	ctx := context.TODO()
 	mockNodesProcessed := 3
 
-	rcRollingUpgrade.finishExecution(upgrademgrv1alpha1.StatusError, mockNodesProcessed, &ctx, ruObj)
+	err = errors.New("execution error")
```
shrinandj, line 975: This could also replaced with `fmt.Errorf`

adammw:
@shrinandj any more blockers to merge?

PR #167, opened 2020-12-18
Move constants to types so that they can be reused

uthark:
Can't reuse from current package because it tries to pull dependency on kubernetes api, so build fails with mismatched dependency versions.
Moving constants to types removes dependency and simplifies reuse  outside.

These constants are part of the public API - they describe status of the rollout, so should be easy to consume.


shrinandj (approved):

uthark:
@shrinandj could you please merge it? I don't have permissions to do it.

PR #168, opened 2020-12-18
Remove separate module for pkg/log

uthark:
Noticed that `pkg/log` lived in a separate module with complicated go.mod setup and it was hard to reuse.
Remove module and make it part of the upgrade manager itself.
Also, in the future may moved retrylogger to a separate package (so that if you want only logger interface, you don't pull dependency on aws sdk).


uthark:
@shrinandj and/or @eytan-avisror could you please take a look?

shrinandj (approved):

PR #170, opened 2020-12-18
Fix namespaced name order

uthark:
Namespaced name should be namespace/name, not the other way.


uthark:
@eytan-avisror / @shrinandj please take a look.

shrinandj (approved):

shrinandj:
If there are more suggestions when Eytan gets to review this, we can incorporate them as a separate change.

PR #171, opened 2020-12-18
Use standard fmt.Errorf to format error message; unify error format

uthark:
* Unify error messages.
* Use standard `fmt.Errorf` which support error wrapping.
* Remove dependency on `github.com/pkg/errors`.
* Make tests less fragile.

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -32,10 +32,8 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/keikoproj/aws-sdk-go-cache/cache"
 	iebackoff "github.com/keikoproj/inverse-exp-backoff"
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	v1errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


uthark:
@keikoproj/upgrade-manager-approvers please take a look.

controllers/rollingupgrade_controller.go
```diff
@@ -319,10 +317,10 @@ func (r *RollingUpgradeReconciler) WaitForTermination(ruObj *upgrademgrv1alpha1.
 	return true, nil
 }
 
-func (r *RollingUpgradeReconciler) GetAutoScalingGroup(rollupName string) (*autoscaling.Group, error) {
-	val, ok := r.ruObjNameToASG.Load(rollupName)
+func (r *RollingUpgradeReconciler) GetAutoScalingGroup(namespacedName string) (*autoscaling.Group, error) {
+	val, ok := r.ruObjNameToASG.Load(namespacedName)
```
shrinandj, line 321: Can you confirm that upgrades were tested after this change? Using the namespaced name in the `ruObjNameToASG` map makes sense. Just that `ruObjNameToASG` gets used at various places and need to ensure that those call sites are not making assumptions about the key in this map.

uthark: The cache is updated with namespaced name: https://github.com/keikoproj/upgrade-manager/blob/97a1d4e4aebbaeed808cc096db936be00b4cbeba/controllers/rollingupgrade_controller.go#L453

I renamed variable name to clearly indicate that is a namespaced name.

shrinandj (approved):

PR #174, opened 2021-01-05
Output can contain other messages from API Server, so be more relaxed when checking output.

uthark:
For example, if FlowControl is used, then calls may be throttled and upgrade manager will incorrectly fail the drain if node is gone.

```
{"level":"error","ts":1609870055.4765916,"logger":"controllers.RollingUpgrade","msg":"Script finished","rollingupgrade":"rollingupgrade-api-us-east-2b","output":"I0105 18:07:27.876891     140 request.go:645] Throttling request took 1.097511969s, request: GET:https://10.236.52.1:443/apis/node.k8s.io/v1beta1?timeout=32s\nError from server (NotFound): nodes \"ip-10-221-247-128.us-east-2.compute.internal\" not found\n","error":"exit status 1","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/keikoproj/upgrade-manager/controllers.(*ScriptRunner).error\n\t/workspace/controllers/script_runner.go:101\ngithub.com/keikoproj/upgrade-manager/controllers.(*ScriptRunner).runScriptWithEnv\n\t/workspace/controllers/script_runner.go:77\ngithub.com/keikoproj/upgrade-manager/controllers.(*ScriptRunner).runScript\n\t/workspace/controllers/script_runner.go:85\ngithub.com/keikoproj/upgrade-manager/controllers.(*ScriptRunner).drainNode\n\t/workspace/controllers/script_runner.go:57\ngithub.com/keikoproj/upgrade-manager/controllers.(*RollingUpgradeReconciler).CallKubectlDrain\n\t/workspace/controllers/rollingupgrade_controller.go:219"}
```

PR #175, opened 2021-01-06
Use kubectl-pkg-provided drain rather than shelling out

adammw:
Potential other fix for #174 - directly run the drain equivalent from upgrade-manager directly rather than shelling out to kubectl.

The k8s.io/kubectl/pkg/drain seems to be rather undocumented/unused so not sure exactly how well it works, but the existence and comments bode well for it to be embedded in other applications.

cc @justinsb / questions to be answered - is this the expected/correct invocation of this pkg? are there any major differences between the `kubectl drain` logic and this method?


controllers/rollingupgrade_controller.go
```diff
@@ -209,13 +211,18 @@ func (r *RollingUpgradeReconciler) DrainNode(ruObj *upgrademgrv1alpha1.RollingUp
 // CallKubectlDrain runs the "kubectl drain" for a given node
 // Node will be terminated even if pod eviction is not completed when the drain timeout is exceeded
 func (r *RollingUpgradeReconciler) CallKubectlDrain(nodeName string, ruObj *upgrademgrv1alpha1.RollingUpgrade, errChan chan error) {
-	out, err := r.ScriptRunner.drainNode(nodeName, ruObj)
```
grosser, line 212: delete the drainNode method and whatever else it calls too ?

uthark:
@adammw Thank you for the PR, this is great improvement.

shreyas-badiger:
I have implemented this in upgrade-manager version 2 (controller v2 branch). Thanks for the suggestion.

PR #178, opened 2021-02-02
Controller v2: Implementation of Instance termination

shreyas-badiger:


controllers/rollingupgrade_controller.go
```diff
@@ -161,3 +166,18 @@ func (r *RollingUpgradeReconciler) UpdateStatus(rollingUpgrade *v1alpha1.Rolling
 		r.Error(err, "failed to update status", "name", rollingUpgrade.NamespacedName())
 	}
 }
+
+func (r *RollingUpgradeReconciler) IsRollingUpgradeCompleted(rollingUpgrade *v1alpha1.RollingUpgrade) bool {
+	r.Info("checking if rolling upgrade is completed", "name", rollingUpgrade.NamespacedName())
+	if err := r.Cloud.Discover(); err != nil {
```
shreyas-badiger, line 172: I am not sure if I need to cache here or wait for the cache to refresh by itself.

tekenstam: I don't think `Discover()` is needed here since it was already called in `Reconcile()` before `IsRollingUpgradeCompleted()` is called. I think the basic structure of `Reconcile()` is:
1. Check if this CR can be admitted for processing
2. Refresh AWS cache by calling `Cloud.Discover()`.
3. Do the work needed based on the state of resources in the AWS cache.
I think you only would need to call `Cloud.Discover()` once.

tekenstam (commented):
Looks good. I think you can remove the extra `Discover()` call from `IsRollingUpgradeCompleted()`.

api/v1alpha1/rollingupgrade_types.go
```diff
@@ -117,9 +116,21 @@ const (
...
+
+//ASG Instance states
+const (
+	InService          = "InService"
+	Pending            = "Pending"
+	PendingWait        = "Pending:Wait"
+	PendingProceed     = "Pending:Proceed"
+	Terminating        = "Terminating"
+	TerminatingWait    = "Terminating:Wait"
+	TerminatingProceed = "Terminating:Proceed"
```
eytan-avisror, line 133: I don't think we need to define this, these should be included in the autoscaling pkg as constants..
see autoscaling.LifecycleState<> constants

controllers/providers/aws/utils.go
```diff
@@ -116,3 +115,17 @@ func GetTemplateLatestVersion(templates []*ec2.LaunchTemplate, templateName stri
 	}
 	return "0"
 }
+
+func TagEC2instance(instanceID, tagKey, tagValue string, client ec2iface.EC2API) error {
```
eytan-avisror, line 119: Can you make this a method of `AmazonClientSet` and place it under ec2.go?
Make sure you use the object instead of taking in a client

shreyas-badiger: Yup. I shall do that.

controllers/rollingupgrade_controller.go
eytan-avisror: I think we can remove these last two blocks all together.
The CR should be marked completed within upgrade.go, once the last instance is rotated and we detect no more instances that require rotation.

Then there is no need for `IsRollingUpgradeCompleted` or the above check of:
```
	if r.IsRollingUpgradeCompleted(rollingUpgrade) {
		rollingUpgrade.SetCurrentStatus(v1alpha1.StatusComplete)
	}
```

This is because on the next reconcile we will finish via:
```
// Stop processing upgrades which are in finite state
	currentStatus := rollingUpgrade.CurrentStatus()
	if common.ContainsEqualFold(v1alpha1.FiniteStates, currentStatus) {
		r.AdmissionMap.Delete(req.NamespacedName)
		r.Info("rolling upgrade ended", "name", req.NamespacedName, "status", currentStatus)
		return reconcile.Result{}, nil
	}
```

Definitely avoid calling `Discover()` more than once, when you need to refresh your state, requeue

api/v1alpha1/rollingupgrade_types.go
shreyas-badiger: Found it. I shall use this one.
https://github.com/aws/aws-sdk-go/blob/4a3fa399d8948450087b360a1d82e8298eb1c25c/service/autoscaling/api.go#L14900

controllers/upgrade.go
```diff
@@ -113,6 +121,44 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
 	return true, nil
 }
 
+func (r *RollingUpgradeReconciler) SetInstanceTag(rollingUpgrade *v1alpha1.RollingUpgrade, instance *autoscaling.Instance, tagKey string, tagVal string) error {
+	return awsprovider.TagEC2instance(aws.StringValue(instance.InstanceId), tagKey, tagVal, r.Auth.Ec2Client)
```
eytan-avisror, line 125: Once you move `TagEC2instance` under the `AmazonClientSet` this call should change to be made through that object

controllers/upgrade.go
```diff
@@ -113,6 +121,44 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
...
+}
+
+func (r *RollingUpgradeReconciler) CanTerminate(instance *autoscaling.Instance) bool {
+	instances, err := r.Auth.DescribeTaggedInstanceIDs(instanceStateTagKey, completedTagValue)
```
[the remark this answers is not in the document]
eytan-avisror: You can technically just use `common.ContainsEqualFold(r.Cloud, "i-123123123")` to know if an instance is in progress, so we can avoid adding another predciate `CanTerminate`

controllers/upgrade.go
```diff
@@ -113,6 +121,44 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
...
+		return false
+	}
+
+	if common.ContainsEqualFold(v1alpha1.ASGNonRunningInstanceStates, aws.StringValue(instance.LifecycleState)) {
+		return false
+	}
+	return true
+}
+
+func (r *RollingUpgradeReconciler) TerminateNodeInstances(rollingUpgrade *v1alpha1.RollingUpgrade, instance *autoscaling.Instance) bool {
```
eytan-avisror, line 146: Please move this under `AmazonClientSet` object under `autoscaling.go`.

controllers/upgrade.go
```diff
@@ -144,6 +190,9 @@ func (r *RollingUpgradeReconciler) SelectTargets(rollingUpgrade *v1alpha1.Rollin
 	if rollingUpgrade.UpdateStrategyType() == v1alpha1.RandomUpdateStrategy {
 		for _, instance := range scalingGroup.Instances {
 			if r.IsInstanceDrifted(rollingUpgrade, instance) {
+				if !r.CanTerminate(instance) {
```
eytan-avisror, line 193: We already handle in progress instances here:
```
for _, instance := range r.Cloud.InProgressInstances {
		selectedInstance := awsprovider.SelectScalingGroupInstance(instance, scalingGroup)
		targets = append(targets, selectedInstance)
	}
```

controllers/upgrade.go
```diff
@@ -113,6 +121,44 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
...
+	instanceID := aws.StringValue(instance.InstanceId)
+	input := &autoscaling.TerminateInstanceInAutoScalingGroupInput{
+		InstanceId:                     aws.String(instanceID),
+		ShouldDecrementDesiredCapacity: aws.Bool(false),
+	}
+
+	r.Info("Terminating node-instance", "instance", instanceID)
+	_, err := r.Auth.AsgClient.TerminateInstanceInAutoScalingGroup(input)
+	if err != nil {
+		r.Error(err, "Failed to terminate instance", "name", rollingUpgrade.NamespacedName(), "instance", instanceID)
```
eytan-avisror, line 156: can you also make sure to return the error instead of logging it here?
Let the caller decide what to do

api/v1alpha1/rollingupgrade_types.go
```diff
@@ -117,9 +117,14 @@ const (
 )
 
 var (
-	FiniteStates        = []string{StatusComplete, StatusError}
-	AllowedStrategyType = []string{string(RandomUpdateStrategy), string(UniformAcrossAzUpdateStrategy)}
-	AllowedStrategyMode = []string{string(UpdateStrategyModeLazy), string(UpdateStrategyModeEager)}
+	FiniteStates                 = []string{StatusComplete, StatusError}
+	AllowedStrategyType          = []string{string(RandomUpdateStrategy), string(UniformAcrossAzUpdateStrategy)}
+	AllowedStrategyMode          = []string{string(UpdateStrategyModeLazy), string(UpdateStrategyModeEager)}
+	ASGTerminatingInstanceStates = []string{
```
eytan-avisror, line 123: nit: can we move this into `aws` package `autoscaling.go` and rename to TerminatingInstanceStates? might be a more appropriate location

controllers/cloud.go
```diff
@@ -29,6 +29,7 @@ import (
 var (
 	instanceStateTagKey = "upgrademgr.keikoproj.io/state"
 	inProgressTagValue  = "in-progress"
+	completedTagValue   = "completed"
```
eytan-avisror, line 32: I don't think we need to set this completed tag.
I think what happens is when we start working on an instance we tag with the `in-progress` tag, and then we know to pick those up first if we resume, however once we terminate, the selector ignores terminated instances, hence we don't really need to make another API call to modify the tag to completed.

controllers/upgrade.go
```diff
@@ -79,6 +89,7 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
 		for _, target := range batch {
 			_ = target
 			// Add in-progress tag
+			r.SetInstanceTag(rollingUpgrade, target, instanceStateTagKey, inProgressTagValue)
```
eytan-avisror, line 92: Can we just call `r.Auth.TagEC2instance` directly and avoid the extra method? we should probably also handle the error condition

controllers/upgrade.go
```diff
@@ -97,6 +108,11 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
 			// Is drained?
 
 			// Terminate - set lastTerminateTime
+			if err := r.Auth.TerminateInstance(target); err != nil {
+				r.Info("Instance termination failed. Will retry in next reconcile", "name", rollingUpgrade.NamespacedName(), "instance", aws.StringValue(target.InstanceId))
```
eytan-avisror, line 112: nit: probably better to log the error, suggest you also simplify the message e.g.
```golang
r.Info("failed to terminate instance", "name", rollingUpgrade.NamespacedName(), "instance", aws.StringValue(target.InstanceId), "error", err)
```
in general I think also more appropriate to keep all log messages lower-case, but that's more personal preference.

controllers/upgrade.go
```diff
@@ -97,6 +108,11 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
...
+				return true, nil
+			}
+			r.SetInstanceTag(rollingUpgrade, target, instanceStateTagKey, completedTagValue)
```
eytan-avisror, line 115: Yeah, so we can probably not do this once we successfully terminate, we don't care about the instance any more, and it would be a waste of API call

controllers/upgrade.go
```diff
@@ -144,6 +164,9 @@ func (r *RollingUpgradeReconciler) SelectTargets(rollingUpgrade *v1alpha1.Rollin
 	if rollingUpgrade.UpdateStrategyType() == v1alpha1.RandomUpdateStrategy {
 		for _, instance := range scalingGroup.Instances {
 			if r.IsInstanceDrifted(rollingUpgrade, instance) {
+				if common.ContainsEqualFold(v1alpha1.ASGTerminatingInstanceStates, aws.StringValue(instance.LifecycleState)) {
```
eytan-avisror, line 167: Should we include this if condition within `IsInstanceDrifted`?

api/v1alpha1/rollingupgrade_types.go
shreyas-badiger: You are right. Let me do that.

controllers/cloud.go
shreyas-badiger: I agree. This was before I started filtering out instances based on states.

controllers/upgrade.go
shreyas-badiger: I first tried logging the error. The entire console becomes too messy with the long error messages. But the rollup succeeds anyway. I thought these long error messages could mislead the observer. WDYT?

Regarding the case, let's keep it uniform. I will keep all the logging in lowercase.

controllers/upgrade.go
shreyas-badiger: I agree. Removing this.

controllers/upgrade.go
shreyas-badiger: Haha! I had the same thoughts too. I see no harm in doing so. I think I can move it there.

controllers/upgrade.go
eytan-avisror: Well, we need to know why the call failed.. I think the error should not be too long, can you give an example to how the log was messy? did it add a stack trace?

shreyas-badiger: Yup. It adds stacktrace.

tekenstam (approved):

controllers/upgrade.go
```diff
@@ -79,6 +89,7 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
 		for _, target := range batch {
 			_ = target
 			// Add in-progress tag
+			r.Auth.TagEC2instance(aws.StringValue(target.InstanceId), instanceStateTagKey, inProgressTagValue)
```
eytan-avisror, line 92: Do we want to handle the error here if it happens?

controllers/upgrade.go
shreyas-badiger: I have added the error log messages for now. You are right, we should know why it failed.
I think we can do a final review of all the logging that goes to the console as a ticket.

eytan-avisror (approved):
👌

PR #181, opened 2021-02-11
Implemented node drain.

shreyas-badiger:
Unlike the current version of the controller, in V2 we will be calling drain using the kubectl-pkg-provided-drain rather than calling the drain using `kubectl` binary.


controllers/providers/kubernetes/nodes.go
```diff
@@ -32,3 +38,45 @@ func (k *KubernetesClientSet) ListClusterNodes() (*corev1.NodeList, error) {
...
+		GracePeriodSeconds:  -1,
+		IgnoreAllDaemonSets: true,
+		Out:                 os.Stdout,
+		ErrOut:              os.Stdout,
+		DeleteEmptyDirData:  true,
+		Timeout:             time.Duration(120) * time.Second,
+	}
+
+	if err := drain.RunCordonOrUncordon(helper, node, true); err != nil {
+		if apierrors.IsNotFound(err) {
```
eytan-avisror, line 64: Did you verify this returns an IsNotFound error?

controllers/providers/kubernetes/nodes.go
```diff
@@ -32,3 +38,45 @@ func (k *KubernetesClientSet) ListClusterNodes() (*corev1.NodeList, error) {
...
+	}
+
+	if err := drain.RunNodeDrain(helper, node.Name); err != nil {
+		if apierrors.IsNotFound(err) {
+			return err
+		}
+		return fmt.Errorf("error draining node: %v", err)
+	}
+
+	if PostDrainDelaySeconds > 0 {
```
eytan-avisror, line 77: This should be done via requeue and comparison of timestamp - no more sleeping :)

controllers/providers/kubernetes/nodes.go
```diff
@@ -32,3 +38,45 @@ func (k *KubernetesClientSet) ListClusterNodes() (*corev1.NodeList, error) {
...
+
+	helper := &drain.Helper{
+		Client:              client,
+		Force:               true,
+		GracePeriodSeconds:  -1,
+		IgnoreAllDaemonSets: true,
+		Out:                 os.Stdout,
+		ErrOut:              os.Stdout,
+		DeleteEmptyDirData:  true,
+		Timeout:             time.Duration(120) * time.Second,
```
eytan-avisror, line 60: What is the timeout we set in v1 controller? can we keep the same?

controllers/providers/kubernetes/nodes.go
eytan-avisror: We can do this in a separate story where we handle parallelizing this

shreyas-badiger: Ah! Yeah. I think we are already doing that as I am setting the lastNodeDrainTime. I shall remove this sleep.

controllers/providers/kubernetes/nodes.go
shreyas-badiger: Yes. I have to pull that from Spec. I just did

controllers/providers/kubernetes/nodes.go
shreyas-badiger: So, under apierrors, there are several different type of errors. IsNotFound() returns error when a resource type is not found.
"IsNotFound returns true if the specified error was created by NewNotFound. NewNotFound returns a new error which indicates that the resource of the kind and the name was not found."

However, I wasn't able to catch the error for invalid nodename. I checked the code for RunNodeDrain(),  they do return an error when they can't select pods for eviction, but I didn't see any check for empty list of pods. I mean, when the pod selector returns an empty list, they just move on to the next steps.

tekenstam (approved):

PR #183, opened 2021-02-17
Eager mode implementation

shreyas-badiger:
- Setting instance to StandBy mode.
- Wait for the desired instances/nodes

### DIFF
--- a/controllers/upgrade.go
+++ b/controllers/upgrade.go
@@ -115,9 +115,12 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
 			}
 
 			// Standby
-			if err := r.SetInstanceStandBy(target, rollingUpgrade); err != nil {
-				r.Error(err, "couldn't set the instance to stand-by", "name", rollingUpgrade.NamespacedName(), "instance", instanceID)
-				return false, err
+			if aws.StringValue(target.LifecycleState) == autoscaling.LifecycleStateInService {
+				r.Info("setting instance to stand-by", "name", rollingUpgrade.NamespacedName(), "instance", instanceID)
+				if err := r.Auth.SetInstanceStandBy(target, rollingUpgrade.Spec.AsgName); err != nil {
+					r.Error(err, "couldn't set the instance to stand-by", "name", rollingUpgrade.NamespacedName(), "instance", instanceID)
+					return false, err
+				}
 			}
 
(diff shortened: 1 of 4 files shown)


controllers/upgrade.go
```diff
@@ -111,13 +111,21 @@ func (r *RollingUpgradeReconciler) ReplaceNodeBatch(rollingUpgrade *v1alpha1.Rol
...
 			}
 
 			// Standby
+			if err := r.SetInstanceStandBy(target, rollingUpgrade); err != nil {
+				r.Error(err, "couldn't set the instance to stand-by", "name", rollingUpgrade.NamespacedName(), "instance", instanceID)
+				return false, err
+			}
 
 			// Wait for desired nodes
+			if !r.DesiredNodesReady(rollingUpgrade) {
```
eytan-avisror, line 124: Should we put a log line here instead of the internal logging?

controllers/upgrade.go
```diff
@@ -334,3 +337,56 @@ func (r *RollingUpgradeReconciler) IsScalingGroupDrifted(rollingUpgrade *v1alpha
...
+	)
+
+	if instanceState != autoscaling.LifecycleStateInService {
+		return nil
+	}
+	r.Info("setting instance to stand-by", "name", rollingUpgrade.NamespacedName(), "instance", instanceID)
+
+	input := &autoscaling.EnterStandbyInput{
+		AutoScalingGroupName:           aws.String(rollingUpgrade.Spec.AsgName),
+		InstanceIds:                    aws.StringSlice([]string{instanceID}),
```
eytan-avisror, line 354: Should we do this for the whole batch? i.e. take multiple instances/ids as arguments

eytan-avisror: Also, should we have a wrapper for the AWS call of EnterStandby inside autoscaling.go?

controllers/upgrade.go
shreyas-badiger: You mean, instead of error logging?

controllers/upgrade.go
shreyas-badiger: Yes, we should move it to autoscaling.go
Regarding sending the whole batch to standby, I think it will get addressed in the next ticket anyway. (parallelize the batch execution). Is that alright? We can also do that for terminate as well.

eytan-avisror (approved):
LG

tekenstam (commented):
LGTM

PR #185, opened 2021-02-20
Fail the CR for drain failures, when IgnoreDrainFailures isn't set.

shreyas-badiger:
- [x] Functional test

### DIFF
--- a/controllers/rollingupgrade_controller_test.go
+++ b/controllers/rollingupgrade_controller_test.go
@@ -2766,6 +2766,36 @@ func TestDrainNodeTerminateTerminatesWhenIgnoreDrainFailuresSet(t *testing.T) {
 
 }
 
+func TestPreDrainFailureWhenIgnoreDrainFailuresSet(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+	mockNode := "some-node-name"
+
+	mockKubeCtlCall := "exit 1;"
+
+	ruObj := &upgrademgrv1alpha1.RollingUpgrade{
+		ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"},
+		Spec: upgrademgrv1alpha1.RollingUpgradeSpec{
+			Strategy:            upgrademgrv1alpha1.UpdateStrategy{DrainTimeout: -1},
+			IgnoreDrainFailures: true,
+			PreDrain: upgrademgrv1alpha1.PreDrainSpec{
+				Script: mockKubeCtlCall,
+			},
+		},
+	}
+	rcRollingUpgrade := &RollingUpgradeReconciler{
+		ClusterState: NewClusterState(),
+		Log:          log2.NullLogger{},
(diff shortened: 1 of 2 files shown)


controllers/rollingupgrade_controller_test.go
```diff
@@ -2730,8 +2734,7 @@ func TestDrainNodeTerminateTerminatesWhenIgnoreDrainFailuresSet(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
 	mockNode := "some-node-name"
 
-	// Force quit from the rest of the command
-	mockKubeCtlCall := "exit 1;"
+	mockKubeCtlCall := ""
```
kevdowney, line 2737: Why is there no more exit 1 if the test case here is meant to simulate error?

shreyas-badiger: Actually, I am not sure if this test case was written to simulate the error or test if the `IgnoreDrainFailuresSet` goes through, because they were checking the error to be nil? If it's former, then I should check the error to be not nil.

shreyas-badiger: The comment specifically says don't expect errors.
That's why I had to get rid of exit and put a empty command. I can even put a sample command as "echo xyz" but again that would be the same. 

`g.Expect(err).To(gomega.BeNil()) // don't expect errors.`

kevdowney: I think the test is saying don't expect errors when any error happens in script, which it expects would include a drain? Not sure how you will know if there's an actual kubectl call in the script unless that's being parsed.

Anyhow, if you are setting `IgnoreDrainFailures: true` won't we need to simulate a failure?

eytan-avisror: I think this test is meant to test that preDrainHelper (which is called as part of drain) is also included in the logic of `IgnoreDrainFailuresSet` - e.g., if the predrain script fails (and drain is then skipped) the test asserts that failure should still be skipped if IgnoreDrainFailuresSet is true (expect no errors even though it exit 1)

This is actually bad logic - if the predrain script fails  - `IgnoreDrainFailuresSet` should not matter, since it's not a drain failure, it's a script execution failure - this should in my mind fail the upgrade entirely.

So I  think Shreyas' change addresses this as well .

eytan-avisror (approved):
LGTM

controllers/rollingupgrade_controller_test.go
shreyas-badiger: @eytan-avisror That's right. Earlier the Predrain failure would also be ignored if `IgnoreDrainFailures`was set. Now it won't. 

@kevdowney I get your point. I will add another unit test to simulate the preDrain failure(with `exit 1;`) while `IgnoreDrainFailures` is set. As Eytan explained, we should expect an error in this case. :)

kevdowney (approved):

PR #192, opened 2021-03-09
Process the batch rotation in parallel

shreyas-badiger:
- [x] Functional tests.
- [x] Minor bug fixes for other functions.

### DIFF
--- a/controllers/upgrade.go
+++ b/controllers/upgrade.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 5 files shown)


controllers/upgrade.go
```diff
@@ -296,25 +354,25 @@ func (r *RollingUpgradeReconciler) IsInstanceDrifted(rollingUpgrade *v1alpha1.Ro
 			upgradeCreationTime = rollingUpgrade.CreationTimestamp.Time
 		)
 		if nodeCreationTime.Before(upgradeCreationTime) {
-			r.Info("rolling upgrade configured for forced refresh", "name", rollingUpgrade.NamespacedName(), "instance", instanceID)
+			r.Info("rolling upgrade configured for forced refresh", "instance", instanceID, "name", rollingUpgrade.NamespacedName())
```
shaoxt, line 357: Why do we switch the order for instanceID and namespace ?
I think it's better to have instance Name here instread of InstanceID.

shaoxt (approved):
LGTM

controllers/upgrade.go
shreyas-badiger: I did that for the sake of uniformity. In instance manager, old controller, this was the order followed. Namespace name comes in the end.

PR #204, opened 2021-03-30
new nodes are only ready if they have the expected launch template

grosser:
before terminating a node we need to wait for a new node with the correct launch template to exist

before it saw that the number of nodes matches and started draining
after we wait for new nodes to come up before draining old ones


grosser:
this does not work because it would wait forever when updating more than 1 instance since the asg will never have all instances on the right version ...

PR #205, opened 2021-03-30
when putting an instance in standby the asgs cache needs to be expired

grosser:
... otherwise when checking the asg in WaitForDesiredInstances it looks like it has enough InService instances already

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -248,7 +248,7 @@ func (r *RollingUpgradeReconciler) WaitForDesiredInstances(ruObj *upgrademgrv1al
 	var err error
 	var ieb *iebackoff.IEBackoff
 	for ieb, err = iebackoff.NewIEBackoff(WaiterMaxDelay, WaiterMinDelay, 0.5, WaiterMaxAttempts); err == nil; err = ieb.Next() {
-		err = r.populateAsg(ruObj)
+		err = r.populateAsg(ruObj, false)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


eytan-avisror:
Can you describe the issue you faced, and how to reproduce it?

shreyas-badiger:
I think invalidating the cache here totally makes sense as we just updated the ASG.

grosser:
issue was that the rollout completed even though some instances were never rolled, leading to old nodes hanging around

grosser:
updated it to populate instead of clearing the cache, since clearing the cache just makes cache lookups crash (maybe that should not be a public api or the cache should know how to refresh, but that's a bigger refactor)

grosser:
@shreyas-badiger  looks good ?

controllers/rollingupgrade_controller.go
```diff
@@ -744,7 +751,7 @@ func (r *RollingUpgradeReconciler) Process(ctx *context.Context,
 func (r *RollingUpgradeReconciler) validateNodesLaunchDefinition(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
 	//Get ASG launch config
 	var err error
-	err = r.populateAsg(ruObj)
+	err = r.populateAsg(ruObj, false)
```
shreyas-badiger, line 754: This has to be `true` as we are doing one final check before marking the rolling upgrade CR completed.

controllers/rollingupgrade_controller.go
```diff
@@ -458,9 +465,9 @@ func (r *RollingUpgradeReconciler) getNodeFromAsg(i *autoscaling.Instance, nodeL
 	return nil
 }
 
-func (r *RollingUpgradeReconciler) populateAsg(ruObj *upgrademgrv1alpha1.RollingUpgrade) error {
+func (r *RollingUpgradeReconciler) populateAsg(ruObj *upgrademgrv1alpha1.RollingUpgrade, force bool) error {
```
shreyas-badiger, line 468: Please remove this change. We added the populateAsg() at multiple cases to handle the edge cases. This means, we have to cache at that point in time else we see failures in upgrades. So, all the  calls for populateAsg() will end up having a ` true`. 
So, for now just call a populateAsg() as soon as we put the instance to standby. 

PS: There is a bigger refactor already in place to handle such in situations in upgrade-manager-v2

PR #206, opened 2021-03-31
(WIP) Unit tests

shreyas-badiger:


controllers/rollingupgrade_controller.go
```diff
@@ -73,7 +73,7 @@ func (r *RollingUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	err := r.Get(ctx, req.NamespacedName, rollingUpgrade)
 	if err != nil {
 		if kerrors.IsNotFound(err) {
-			r.AdmissionMap.Delete(req.NamespacedName)
+			r.AdmissionMap.Delete(rollingUpgrade.NamespacedName())
```
eytan-avisror, line 76: This will be probematic, if `IsNotFound` is true rollingUpgrade will be nil, best keep it `req.NamespacedName`

eytan-avisror: Downstream from this, it's OK to use `rollingUpgrade.NamespacedName()`

PR #212, opened 2021-04-07
Cache the ASG before nodes are rotated in a loop

shreyas-badiger:
- [x] Manual testing
- [x] Edge cases

There is a bug in the node rotation in upgrade-manager-v1 that will fail all the rollup CRs that have `StandBy/in-progress` instances

In the current implementation of upgrade-manager-v1, the node rotation mainly happens [in this loop](https://github.com/keikoproj/upgrade-manager/blob/dc2d58bcea4a6dc4202d2559111ef3a1516c46f5/controllers/rollingupgrade_controller.go#L532):

```
asg = getASG() // just loads the ASG and doesn't get the most frequent changes.
totalNodes = asg.Instances
for processedInstances < totalNodes {
    ...
    if thereAreAnyInProgressInstances{
        instances = selectInProgressInstances()
    } else {
        instances = selectInServiceInstances()
    }
rotateInstances(instances)
processedInstances += len(instances)
}
```

The final line **processedInstances += len(instances)** accounts not just the `InService` instances but also the `StandBy/ in-progress` instances.

Let's say, the ASG had 3 `InService` instances and one instance in `StandBy/in-progress`. So, we had 4 instances in total.
But the ASG was cached previously and had a count of 3 in `asgSize`, hence we skipped one instance because of which the launch config validation failed.

`2021-04-05T12:13:22.619Z	INFO	controllers.RollingUpgrade	Nodes in ASG that *might* need to be updated	{"rollingupgrade": "addon-upgrademgr-ns/iks-arktika-bdd-nightly-usw2-system-ng-nodelaunchconfig-qoq8zwub09zx", "asgName": "iks-arktika-bdd-nightly-usw2-system-ng-NodeGroup-1WVN49BGRKUMU", "asgSize": 3}`

In sum, the rollup CR will mostly fail whenever the upgrade-manager picks up a `StandBy/ in-progress` instance.  

**Quick fix: Cache the ASG just before we start the rotation.** 
```
asg = populateASG() // refresh AWS cache
totalNodes = asg.Instances
for processedInstances < totalNodes {
    ...
    if thereAreAnyInProgressInstances{
        instances = selectInProgressInstances()
    } else {
        instances = selectInServiceInstances()
    }
rotateInstances(instances)
processedInstances += len(instances)
}
```

PS: This is a bad design and costly as we are doing yet another AWS caching. This will be addressed in the upgrade-manager-v2 as we will cache only once at the beginning of the reconcile.

### DIFF
--- a/controllers/rollingupgrade_controller_test.go
+++ b/controllers/rollingupgrade_controller_test.go
@@ -1138,28 +1138,6 @@ func TestRunRestackSameLaunchConfig(t *testing.T) {
 	g.Expect(err).To(gomega.BeNil())
 }
 
-func TestRunRestackRollingUpgradeNotInMap(t *testing.T) {
-	g := gomega.NewGomegaWithT(t)
-
-	strategy := upgrademgrv1alpha1.UpdateStrategy{Mode: upgrademgrv1alpha1.UpdateStrategyModeLazy}
-	ruObj := &upgrademgrv1alpha1.RollingUpgrade{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"},
-		Spec: upgrademgrv1alpha1.RollingUpgradeSpec{Strategy: strategy},
-	}
-	rcRollingUpgrade := &RollingUpgradeReconciler{
-		Log:          log2.NullLogger{},
-		ClusterState: NewClusterState(),
-		ASGClient:    MockAutoscalingGroup{},
-		EC2Client:    MockEC2{},
-	}
-	ctx := context.TODO()
-
-	g.Expect(rcRollingUpgrade.ruObjNameToASG.Load(ruObj.NamespacedName())).To(gomega.BeNil())
-	int, err := rcRollingUpgrade.runRestack(&ctx, ruObj)
-	g.Expect(int).To(gomega.Equal(0))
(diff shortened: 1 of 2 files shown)


kevdowney (approved):
LGTM

controllers/rollingupgrade_controller_test.go
```diff
@@ -1138,28 +1138,6 @@ func TestRunRestackSameLaunchConfig(t *testing.T) {
 	g.Expect(err).To(gomega.BeNil())
 }
 
-func TestRunRestackRollingUpgradeNotInMap(t *testing.T) {
```
eytan-avisror, line 1141: Any particular reason we are removing this test?

shreyas-badiger: This test assumes that runRestack won't be adding this rollingUpgrade object to the sync map.
But by calling the populateAsg(ruObj) we will also be adding the rollingUpgrade object to the sync map. populateAsg() calls Store function. Hence the purpose of this test is defeated.

shreyas-badiger: Screen Shot 2021-04-13 at 9 31 52 AM

eytan-avisror (approved):

PR #216, opened 2021-04-16
Issue 2108: step duration metrics to v1

sahilbadla:
Related to issue https://github.intuit.com/kubernetes/arktika/issues/2108

Add step duration metrics to v1(followed similar change from v2)
Added service to expose metrics for prometheus


shaoxt (approved):
LGTM

eytan-avisror (approved):
👍

PR #218, opened 2021-04-21
#2285: rollup CR statistic metrics in v2

sahilbadla:
Signed-off-by: sbadla1 <sahil_badla@intuit.com>

### DIFF
--- a/controllers/common/metrics.go
+++ b/controllers/common/metrics.go
@@ -11,11 +11,13 @@ import (
 )
 
 var (
+	metricNamespace = "upgrade_manager_v2"
+
 	//All cluster level node upgrade statistics
 	nodeRotationTotal = prometheus.NewHistogram(
 		prometheus.HistogramOpts{
-			Namespace: "node",
-			Name:      "rotation_total_seconds",
+			Namespace: metricNamespace,
+			Name:      "node_rotation_total_seconds",
 			Help:      "Node rotation total",
 			Buckets: []float64{
 				10.0,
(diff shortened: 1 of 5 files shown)


controllers/common/metrics.go
```diff
@@ -1,38 +1,59 @@
...
-			120.0,
-			180.0,
-			300.0,
-			600.0,
-			900.0,
+	stepSummaries = make(map[string]map[string]prometheus.Summary)
+
+	CRStatus = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "upgrade-manager",
```
eytan-avisror, line 37: can we make sure all metrics are under a namespace `upgrade_manager`? you can make that a const.
That way all controller metrics we publish would show up as `upgrade_manager_<metric-name>`

controllers/common/metrics.go
```diff
@@ -1,38 +1,59 @@
...
+			Subsystem: "v2",
+			Name:      "cr_status",
```
eytan-avisror, line 39: since this is a metric name, can we avoid the abbreviation? e.g. `resource_status` instead

controllers/common/metrics.go
```diff
@@ -1,38 +1,59 @@
...
+			Help:      "Rollup CR statistics, partitioned by name and type.",
+		},
+		[]string{
+			// name of the CR
+			"name",
```
eytan-avisror, line 44: this label might conflict with other labels, suggest you change to `resource_name`

controllers/common/metrics.go
```diff
@@ -1,38 +1,59 @@
...
+			// status stype of CR, currently only recording "completed" and "failed"
+			"type",
```
eytan-avisror, line 46: this label might conflict with other labels, suggest you change to `resource_status`

controllers/common/metrics.go
shaoxt: Make sense.

sahilbadla: Done

controllers/common/metrics.go
sahilbadla: changed

controllers/common/metrics.go
sahilbadla: changed

controllers/common/metrics.go
sahilbadla: changed label name

eytan-avisror (approved):
LG

PR #231, opened 2021-05-06
V2 controller metrics concurrency fix

shaoxt:
Fix the v2 controller metrics concurrency issue


zhatt (commented):
Is there some way the mutex can be in the RollingUpgradeStatus struct instead of passing it around as part of the API?

eytan-avisror:
+1 for embedding the mutex in a struct instead of passing it.
I think this entire implementation should not be part of `RollingUpgradeStatus` TBH
metrics should be part of reconciler and configure status from within the context of a reconcile

shaoxt:
> Is there some way the mutex can be in the RollingUpgradeStatus struct instead of passing it around as part of the API?

That was the first approach I did.  But the generator does not allow it.

shaoxt:
> +1 for embedding the mutex in a struct instead of passing it.
> I think this entire implementation should not be part of `RollingUpgradeStatus` TBH
> metrics should be part of reconciler and configure status from within the context of a reconcile
The metrics are not only shown in metrics but also on the CR.
That's the reason it has part of RollingUpgradeStatus.

Regarding the mutex.  When I put the following struct in CR. Controller-gen doesn't allow it.

```
type NodeStepMap struct {
	mutex *sync.Mutex

	nodeSteps map[string][]NodeStepDuration
}
```

```
# github.com/keikoproj/upgrade-manager/api/v1alpha1
api/v1alpha1/zz_generated.deepcopy.go:90:11: assignment copies lock value to **out: sync.Mutex
make: *** [vet] Error 2
```

sahilbadla:
@shaoxt is [Sync Map ](https://golang.org/pkg/sync/)a better option here then?

eytan-avisror:
@shaoxt this is because you are putting the implementation inside the API (status) and it gets copied by `zz_generated.deepcopy` - this is why I suggested to move the entire metrics implementation into the context of a RollingUpgrade, and not part of the CR API Objects itself.

Right now we only have a Controller level object `RollingUpgradeReconciler` which holds state that is common to all CRs - this is why you could not find a good place to put it. @shreyas-badiger has a refactor that adds another level `RollingUpgradeContext` which holds state for the individual resource - this would be an idea place to store metrics.

Can you wait for him to merge that PR first? I asked him to send it. It would then be much easier to move this code and embed a mutex.

eytan-avisror:
> @shaoxt is [Sync Map ](https://golang.org/pkg/sync/)a better option here then?

Sync map would work (considering you can copy it - otherwise it would get the same error from `zz_generated.deepcopy`), but it would mean we have to change wherever we access it, add casting, etc + iteration becomes a pain vs. simple as a map with mutex.

shaoxt:
> > @shaoxt is [Sync Map ](https://golang.org/pkg/sync/)a better option here then?
> 
> Sync map would work (considering you can copy it - otherwise it would get the same error from `zz_generated.deepcopy`), but it would mean we have to change wherever we access it, add casting, etc + iteration becomes a pain vs. simple as a map with mutex.

I think moving the method into RollingUpgradeContext makes sense.   Yes.  I considered the SyncMap at the beginning. But have to change all the argument type.

eytan-avisror:
It's merged now @shaoxt if you want to move the implementation

shaoxt:
@eytan-avisror Refined the code based on RollingUpgradeContext

controllers/upgrade.go
```diff
@@ -109,6 +111,110 @@ func (r *RollingUpgradeContext) RotateNodes() error {
 	return nil
 }
 
+// Update last batch nodes
+func (s *RollingUpgradeContext) UpdateLastBatchNodes(batchNodes map[string]*v1alpha1.NodeInProcessing) {
```
eytan-avisror, line 115: Can we move all these methods to `controllers/metrics.go`?

controllers/upgrade_test.go
```diff
@@ -385,3 +386,48 @@ func TestDesiredNodesReady(t *testing.T) {
 		}
 	}
 }
+
+// Test
+func TestNodeTurnsOntoStep(t *testing.T) {
```
eytan-avisror, line 391: Also this to `controllers/metrics_test.go`

controllers/upgrade.go
shaoxt: package "api" depends on package "common". 
We can not let package "common" refer to the structures in "api.*".
Also common is used in "controllers", those methods need to access RollingUpgradeContext.

eytan-avisror: This can stay same package @shaoxt 
just different file

shaoxt: Sounds good to me.  Moved. Thanks

controllers/upgrade_test.go
shaoxt: Moved to upgrade_metrics_test.go

eytan-avisror (approved):

shreyas-badiger (approved):

PR #232, opened 2021-05-07
Fix metrics concurrency issue 

shaoxt:
Fix metrics concurrency issue


sahilbadla:
@shaoxt do you need to do the same refactoring here as you did in v2?

sahilbadla (approved):

shreyas-badiger (approved):

PR #236, opened 2021-05-17
upgrade-manager-v2: Move DrainManager back to Reconciler

shreyas-badiger:


controllers/providers/aws/autoscaling.go
```diff
@@ -54,17 +54,23 @@ func (a *AmazonClientSet) TerminateInstance(instance *autoscaling.Instance) erro
...
+			InstanceIds:                    aws.StringSlice(instanceIDs),
+			ShouldDecrementDesiredCapacity: aws.Bool(false),
+		}
+		if _, err := a.AsgClient.EnterStandby(input); err != nil {
+			return len(instanceIDs), err
+		}
 	}
 
-	return nil
+	return len(instanceIDs), nil
```
eytan-avisror, line 75: would a better pattern be returning (bool, error) ?
Do you actually need the number, or do you just need to know if instances were set to standby?

controllers/providers/aws/ec2.go
```diff
@@ -62,16 +63,25 @@ func (a *AmazonClientSet) DescribeTaggedInstanceIDs(tagKey, tagValue string) ([]
...
-		Resources: aws.StringSlice([]string{instanceID}),
-		Tags: []*ec2.Tag{
-			{
-				Key:   aws.String(tagKey),
-				Value: aws.String(tagValue),
+func (a *AmazonClientSet) TagEC2instance(instances []*autoscaling.Instance, tagKey, tagValue string) error {
+	var instanceIDs []string
+	for _, instance := range instances {
+		if aws.StringValue(instance.LifecycleState) == autoscaling.LifecycleStateInService {
+			instanceIDs = append(instanceIDs, aws.StringValue(instance.InstanceId))
```
eytan-avisror, line 70: Why do we only want to tag InService instances? you can tag instances regardless of their state, no?

controllers/providers/aws/autoscaling.go
shreyas-badiger: This makes sense. I can use bool

controllers/providers/aws/ec2.go
shreyas-badiger: The possible states the instance can be when it arrives at this point in code are - `InService, Pending, Detaching, StandBy`. Instances that are in `StandBy` already have in-progress tag.  By this check, we can avoid the additional AWS calls to tag instances repeatedly or avoid tagging the instances that we aren't ready to operate on.

eytan-avisror: Yeah, but it doesn't make sense from code perspective, because the method `TagEC2instance` is supposed to be generic, if we want to do any filtering based on lifecycle-state it should happen on the controller side e.g. only pass InService instances to TagEC2Instance - you should also rename it to `TagEC2Instances`.

controllers/providers/aws/autoscaling.go
```diff
@@ -54,17 +54,23 @@ func (a *AmazonClientSet) TerminateInstance(instance *autoscaling.Instance) erro
...
+			InstanceIds:                    aws.StringSlice(instanceIDs),
+			ShouldDecrementDesiredCapacity: aws.Bool(false),
+		}
+		if _, err := a.AsgClient.EnterStandby(input); err != nil {
+			return true, err
+		}
 	}
 
-	return nil
+	return false, nil
```
eytan-avisror, line 75: shouldn't the condition be switched? e.g. true = an instance was set to standby, false = there is an error or no instances were set

shreyas-badiger: Oops. Thanks for catching it.

controllers/providers/aws/ec2.go
shreyas-badiger: Hmm. That's right. I can move the checking to controller code.

controllers/providers/aws/autoscaling.go
shreyas-badiger: My bad. It should be this:

```
func (a *AmazonClientSet) SetInstanceStandBy(instances []*autoscaling.Instance, scalingGroupName string) (bool, error) {
	var instanceIDs []string
	for _, instance := range instances {
		if aws.StringValue(instance.LifecycleState) == autoscaling.LifecycleStateInService {
			instanceIDs = append(instanceIDs, aws.StringValue(instance.InstanceId))
		}
	}
	if len(instanceIDs) > 0 {
		input := &autoscaling.EnterStandbyInput{
			AutoScalingGroupName:           aws.String(scalingGroupName),
			InstanceIds:                    aws.StringSlice(instanceIDs),
			ShouldDecrementDesiredCapacity: aws.Bool(false),
		}
		if _, err := a.AsgClient.EnterStandby(input); err != nil {
			return false, err
		}
		return true, nil
	}

	return false, nil
}
```

shreyas-badiger: There could be 3 return conditions.

controllers/upgrade.go
```diff
@@ -140,32 +130,38 @@ func (r *RollingUpgradeContext) ReplaceNodeBatch(batch []*autoscaling.Instance)
...
-			if aws.StringValue(target.LifecycleState) == autoscaling.LifecycleStateInService {
-				r.Info("setting instance to stand-by", "instance", instanceID, "name", r.RollingUpgrade.NamespacedName())
-				if err := r.Auth.SetInstanceStandBy(target, r.RollingUpgrade.Spec.AsgName); err != nil {
-					// failure to set instance to standby are retryable
-					r.Info("failed to set instance to stand-by", "instance", instanceID, "message", err.Error(), "name", r.RollingUpgrade.NamespacedName())
-					return true, nil
-				}
+		// Standby (requeue if any instance was set to stand-by)
+		r.Info("setting batch to stand-by", "instances", awsprovider.GetInstanceIDs(batch), "name", r.RollingUpgrade.NamespacedName())
+		if ok, err := r.Auth.SetInstanceStandBy(batch, r.RollingUpgrade.Spec.AsgName); ok {
```
eytan-avisror, line 143: Condition should be `!ok` here

shreyas-badiger: I have changed this now. The setTag and setStandby function have been simplified and the rest of the conditions in the controller code.

controllers/providers/aws/autoscaling.go
```diff
@@ -54,17 +54,12 @@ func (a *AmazonClientSet) TerminateInstance(instance *autoscaling.Instance) erro
 	return nil
 }
 
-func (a *AmazonClientSet) SetInstanceStandBy(instance *autoscaling.Instance, scalingGroupName string) error {
-	instanceID := aws.StringValue(instance.InstanceId)
+func (a *AmazonClientSet) SetInstanceStandBy(instanceIDs []string, scalingGroupName string) error {
```
eytan-avisror, line 57: can we rename to `SetInstancesStandBy` since we allow batching this now

eytan-avisror (approved):
LG

controllers/providers/aws/autoscaling.go
shreyas-badiger: Done.

PR #250, opened 2021-06-11
set maxUnavailable appropriately.

shreyas-badiger:


controllers/random_node_selector.go
```diff
@@ -13,7 +15,7 @@ type RandomNodeSelector struct {
 }
 
 func NewRandomNodeSelector(asg *autoscaling.Group, ruObj *upgrademgrv1alpha1.RollingUpgrade) *RandomNodeSelector {
-	maxUnavailable := getMaxUnavailable(ruObj.Spec.Strategy, len(asg.Instances))
+	maxUnavailable := getMaxUnavailable(ruObj.Spec.Strategy, int(aws.Int64Value(asg.DesiredCapacity)))
```
eytan-avisror, line 18: Isn't this the same? I guess desired is a more stable number?

shreyas-badiger: `len(asg.Instance)` will also consider the standBy instances. This will adversely impact the % caluclation. For % calculation we should only include the Inservice instances, right?

shreyas-badiger:
I had to delete the fork. Opened another PR - https://github.com/keikoproj/upgrade-manager/pull/255

PR #251, opened 2021-06-14
LastNodeTerminationTime & LastNodeDrainTime not set in v0.19

sahilbadla:
LastNodeTerminationTime & LastNodeDrainTime was not set in the v0.19 of upgrade-manager v1.
This fixes the issue.

### DIFF
--- a/controllers/rollingupgrade_controller.go
+++ b/controllers/rollingupgrade_controller.go
@@ -1048,7 +1048,7 @@ func (r *RollingUpgradeReconciler) DrainTerminate(
 		if err := r.DrainNode(ruObj, nodeName, targetInstanceID, ruObj.Spec.Strategy.DrainTimeout, nodeSteps, inProcessingNodes, mutex); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


controllers/rollingupgrade_controller.go
```diff
@@ -1049,13 +1049,15 @@ func (r *RollingUpgradeReconciler) DrainTerminate(
 			return err
 		}
 	}
+	ruObj.Status.LastNodeDrainTime = metav1.Time{Time: time.Now()}
```
shreyas-badiger, line 1052: Please move this inside the if condition. In v1, we don't drain when a node isn't found.

```
// Drain and wait for draining node.
	if nodeName != "" {
		if err := r.DrainNode(ruObj, nodeName, targetInstanceID, ruObj.Spec.Strategy.DrainTimeout, nodeSteps, inProcessingNodes, mutex); err != nil {
			return err
		}
                ruObj.Status.LastNodeDrainTime = metav1.Time{Time: time.Now()}
	}
```

sahilbadla: Fixed

shreyas-badiger (approved):

PR #271, opened 2021-07-01
upgrade-manager-v2:Add watch event for cluster nodes instead of API calls

shreyas-badiger:


controllers/rollingupgrade_controller.go
```diff
@@ -177,3 +189,24 @@ func (r *RollingUpgradeReconciler) UpdateStatus(rollingUpgrade *v1alpha1.Rolling
 		r.Info("failed to update status", "message", err.Error(), "name", rollingUpgrade.NamespacedName())
 	}
 }
+
+func (r *RollingUpgradeReconciler) nodeReconciler(obj client.Object) []ctrl.Request {
```
eytan-avisror, line 193: We should add handling of nodes getting deleted in case we ever need to check if a node is gone.
You can do this potentially by checking if `!ObjectMeta.DeletionTimestamp.IsZero()` which would indicate a deletion event.
If that is the case you can remove it from the sync map

shaoxt (approved):
LGTM

shreyas-badiger:
Closing this PR as we decided to add node eventHandlers instead of watch.

https://github.com/keikoproj/upgrade-manager/pull/272

PR #278, opened 2021-07-29
Controller v2 bdd changes

sahilbadla:
Added BDD changes as a submodule
added bdd github action


test-bdd/main_test.go
```diff
@@ -0,0 +1,98 @@
+package test
```
shreyas-badiger, line 1: Can you make sure we run this test when we do a "make test" from upgrade-manager repo?
If the make target already runs this, ignore this comment.

test-bdd/templates/rolling-upgrade.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: upgrademgr.keikoproj.io/v1alpha1
```
shreyas-badiger, line 1: This template is not consistent with the new CRD. 
You have to make changes by looking into the CRD in upgrade-manager-v2

https://github.com/keikoproj/upgrade-manager/blob/controller-v2/config/crd/bases/upgrademgr.keikoproj.io_rollingupgrades.yaml

PR #298, opened 2021-08-12
fix panic when using MixedInstancesPolicy

backjo:


controllers/upgrade.go
```diff
@@ -503,7 +503,7 @@ func (r *RollingUpgradeContext) IsInstanceDrifted(instance *autoscaling.Instance
 			launchTemplateName      = aws.StringValue(scalingGroup.MixedInstancesPolicy.LaunchTemplate.LaunchTemplateSpecification.LaunchTemplateName)
 			instanceTemplateName    = aws.StringValue(instance.LaunchTemplate.LaunchTemplateName)
 			instanceTemplateVersion = aws.StringValue(instance.LaunchTemplate.Version)
-			templateVersion         = aws.StringValue(scalingGroup.LaunchTemplate.Version)
+			templateVersion         = aws.StringValue(scalingGroup.MixedInstancesPolicy.LaunchTemplate.LaunchTemplateSpecification.Version)
```
eytan-avisror, line 506: We need to check both `scalingGroup.MixedInstancesPolicy.LaunchTemplate.LaunchTemplateSpecification.Version`
 and `scalingGroup.LaunchTemplate.Version`

eytan-avisror: Otherwise, when a launch template does not have a mixedInstancePolicy you will get a nil pointer for templateVersion

eytan-avisror: I think we need to do a nil check first to understand whether it's a standard launch template or launch template with mixed instance policy, then we know which struct to access to get the name / version

eytan-avisror: Oh sorry, I see we already do it above - this change is correct

backjo:
@eytan-avisror @shreyas-badiger any chance we could get that hotfix with this in?

shreyas-badiger:
@backjo yes. we can do a minor release soon. I have another bug to fix. Will do it right after that.

PR #303, opened 2021-08-25
Set Instances to StandBy in batches

shreyas-badiger:


controllers/upgrade.go
```diff
@@ -160,9 +160,10 @@ func (r *RollingUpgradeContext) ReplaceNodeBatch(batch []*autoscaling.Instance)
 			}
 			// Standby
 			r.Info("setting instances to stand-by", "batch", batchInstanceIDs, "instances(InService)", inServiceInstanceIDs, "name", r.RollingUpgrade.NamespacedName())
-			if err := r.Auth.SetInstancesStandBy(inServiceInstanceIDs, r.RollingUpgrade.Spec.AsgName); err != nil {
-				r.Info("failed to set instances to stand-by", "batch", batchInstanceIDs, "instances(InService)", inServiceInstanceIDs, "message", err.Error(), "name", r.RollingUpgrade.NamespacedName())
+			if err := r.SetBatchStandBy(batchInstanceIDs); err != nil {
+				return false, err
```
shaoxt, line 164: Shall we keep the log

controllers/upgrade.go
```diff
@@ -634,3 +634,29 @@ func (r *RollingUpgradeContext) endTimeUpdate() {
...
+	)
+
+	fmt.Println("SetBatchStandBy", "numberOfBatches", numberOfBatches, "pendingInstances", pendingInstances)
+	for i := 0; i < int(numberOfBatches) && pendingInstances > 0 && start < len(instanceIDs); i++ {
+		batchSize := common.IntMin(pendingInstances, awsprovider.InstanceStandByLimit)
+		if batchSize > len(instanceIDs) {
+			return errors.Errorf("unable to set instances to standBy. Out of bounds error. Instances: %v", instanceIDs[start:batchSize])
+		}
+
+		fmt.Println("SetBatchStandBy(for-loop)", "batchSize", batchSize, "start", start)
```
kevdowney, line 653: This a debug statement?

controllers/upgrade.go
```diff
@@ -634,3 +634,29 @@ func (r *RollingUpgradeContext) endTimeUpdate() {
 		common.TotalProcessingTime(r.RollingUpgrade.ScalingGroupName(), totalProcessingTime)
 	}
 }
+
+// AWS API call for setting an instance to StandBy has a limit of 19. Hence we have to call the API in batches.
+func (r *RollingUpgradeContext) SetBatchStandBy(instanceIDs []string) error {
+	var (
+		pendingInstances = len(instanceIDs)
+		numberOfBatches  = math.Ceil(float64(len(instanceIDs)) / float64(awsprovider.InstanceStandByLimit))
```
shaoxt, line 642: This should be round up, right ?
Let's say   65 instances,   limit is 20,  it requires  4 batches

shaoxt: 60 instances requires 3 batches

controllers/upgrade.go
```diff
@@ -634,3 +634,29 @@ func (r *RollingUpgradeContext) endTimeUpdate() {
...
+		start            = 0
+	)
+
+	fmt.Println("SetBatchStandBy", "numberOfBatches", numberOfBatches, "pendingInstances", pendingInstances)
```
kevdowney, line 646: This a debug statement?

controllers/upgrade.go
```diff
@@ -634,3 +634,29 @@ func (r *RollingUpgradeContext) endTimeUpdate() {
...
+	for i := 0; i < int(numberOfBatches) && pendingInstances > 0 && start < len(instanceIDs); i++ {
```
kevdowney, line 647: I'm wondering if this would be more efficient to chunk the slice  first then loop instead of keeping multiple counters in a loop?


Ex.
```
func chunkSlice(slice []int, chunkSize int) [][]int {
	var chunks [][]int
	for i := 0; i < len(slice); i += chunkSize {
		end := i + chunkSize

		// necessary check to avoid slicing beyond
		// slice capacity
		if end > len(slice) {
			end = len(slice)
		}

		chunks = append(chunks, slice[i:end])
	}

	return chunks
}

```

controllers/upgrade.go
shreyas-badiger: Yes. Sorry. Removing it

controllers/upgrade.go
shreyas-badiger: It is logged in the function SetBatchStandBy

PR #307, opened 2021-09-30
controller flags for ignoreDrainFailures and drainTimeout

shreyas-badiger:


eytan-avisror:
LG, can you just add a few test cases that test the default value vs. flag vs. cr?

shreyas-badiger:
> LG, can you just add a few test cases that test the default value vs. flag vs. cr? e.g. if CR value is nil, it should be the default flag option., if CR is value is nil & flag is provided, it should be flag value, if CR is not nil, it should be the CR value

Yes. It will be hard to see which got enforced. Let me try anyway.

shaoxt (approved):
LGTM

PR #318, opened 2022-02-21
Golang lint fix: remove import duplicates

figo:
using tool https://staticcheck.io/

### DIFF
--- a/api/v1alpha1/rollingupgrade_types_test.go
+++ b/api/v1alpha1/rollingupgrade_types_test.go
@@ -19,9 +19,8 @@ import (
 	"sync"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


eytan-avisror:
@figo 
We don't merge into master any more, this is actually deprecated code, latest code we use is in controller-v2 branch, we should merge it into master but @shreyas-badiger has been having some issues with that.

figo:
> @figo We don't merge into master any more, this is actually deprecated code, latest code we use is in controller-v2 branch, we should merge it into master but @shreyas-badiger has been having some issues with that.

Thanks @eytan-avisror, will close this one. hopefully master branch can be up to date soon.

PR #327, opened 2022-04-15
[Draft PR] "max-replacement-nodes" feature to avoid cluster ballooning during upgrades.

shreyas-badiger:


shreyas-badiger:
Unable to convert it to a PR as Github is including 109 files in diff.

PR #328, opened 2022-04-18
"max-replacement-nodes" feature to avoid cluster ballooning during upgrades.

shreyas-badiger:
`max-replacement-nodes` flag in upgrade-manager limits the number of replacement nodes we can bring up during the cluster upgrade.

This feature helps us limit cluster-ballooning in case of really large clusters and enables us to set `maxUnavailable` to a higher value.

- [x] Functional tests
- [x] Unit tests
- [x] 100 nodes test
- [x] 300 nodes test.


controllers/rollingupgrade_controller.go
```diff
@@ -55,6 +55,8 @@ type RollingUpgradeReconciler struct {
 	ReconcileMap        *sync.Map
 	DrainTimeout        int
 	IgnoreDrainFailures bool
+	ReplacementNodesMap sync.Map
```
eytan-avisror, line 58: Should this be a pointer?
Also, should use just pass the pointer to RollingUpgradeContext instead of entire reconciler?

controllers/upgrade.go
```diff
@@ -662,3 +677,18 @@ func (r *RollingUpgradeContext) SetBatchStandBy(instanceIDs []string) error {
 	}
 	return nil
 }
+
+// Checks for how many replacement nodes exists across all the IGs in the cluster
+func (r *RollingUpgradeContext) ClusterBallooning(batchSize int) bool {
```
eytan-avisror, line 682: Can we add some unit tests around this?

main.go
```diff
@@ -100,6 +101,7 @@ func main() {
 	flag.IntVar(&maxAPIRetries, "max-api-retries", 12, "The number of maximum retries for failed/rate limited AWS API calls")
 	flag.IntVar(&drainTimeout, "drain-timeout", 900, "when the drain command should timeout")
 	flag.BoolVar(&ignoreDrainFailures, "ignore-drain-failures", false, "proceed with instance termination despite drain failures.")
+	flag.IntVar(&maxReplacementNodes, "max-replacement-nodes", 5, "The max number of replacement nodes allowed in a cluster. Avoids cluster-ballooning")
```
eytan-avisror, line 104: Is it a good idea to set the default to 5?
Default should be 0 (no limit) otherwise people would have a breaking change

shreyas-badiger: Oops. That was a default I used during testing. Yes. The idea is to set the default to -1

eytan-avisror: Should we use 0 instead of -1 to indicate no limit? this would align better with other options such as drain-timeout, etc. and also 0 is not a valid option

shreyas-badiger: Sure. I can set it to 0, to keep it uniform.

controllers/rollingupgrade_controller.go
shreyas-badiger: Yes. That makes sense.

shreyas-badiger: @eytan-avisror 
Looks like it isn't allowed.
Passing on `sync.Map` as a pointer.

```
go vet ./...
# github.com/keikoproj/upgrade-manager/controllers
controllers/rollingupgrade_controller.go:177:24: literal copies lock value from r.ReplacementNodesMap: sync.Map contains sync.Mutex
controllers/helpers_test.go:77:24: literal copies lock value from r.ReplacementNodesMap: sync.Map contains sync.Mutex
make: *** [vet] Error 2
```

eytan-avisror: Can you just do a similar implementation to that of the DrainGroup?

controllers/upgrade.go
```diff
@@ -62,6 +62,9 @@ type RollingUpgradeContext struct {
 	metricsMutex        *sync.Mutex
 	DrainTimeout        int
 	IgnoreDrainFailures bool
+	ReplacementNodesMap *sync.Map
+	MaxReplacementNodes int
```
kevdowney, line 66: So do we allow users to override max replacement nodes per spec?

shreyas-badiger: No, we don't. It will remain to be a controller-level flag. Not all the fields in the RollingUpgrade context will end up being a spec in RollingUpgrade object.

kevdowney (approved):
LGTM

vgunapati (approved):
LGTM

PR #351, opened 2023-06-06
only allow maxParallel number of running CRs

ZihanJiang96:
**problem** 
currently `max-parallel` only controls the number of concurrent reconcile() func, in a short period of time, all the IGs will enter the reconcile() at least once, and start to spin up new nodes / drain the old nodes. This will cause a memory spike in upgrade-manager.

**proposal** 
`max-parallel` should limit the max number of RollingUpgrade CRs in `running` state.

**changes introduced by this PR**
1. make sure there is at most `max-parallel` number of IGs in the `AdmissionMap`.
2. if length of `AdmissionMap` is zero, then possibly the pod restarted, fetch all CRs in running state by label, add to admission map, continue the reconcilation against these running IGs.

**tests**

1. unit test pass.
2. tested on a cluster with 50 IG, each IG has 3 node. submit 50 rollingupgrade CR at the same time.
the current implementation requires 1.2GB memory
by limiting `max-parallel` to 10, the memory consumption is 310MB
3. tested on a cluster with 100 IG, each IG has 6 node. submit 100 rollingupgrade CR at the same time.
the current implementation requires 5GB memory, the upgrade time is > 60min
by limiting `max-parallel` to 10, the memory consumption is 700MB, the upgrade time is 85min


controllers/rollingupgrade_controller.go
```diff
@@ -139,6 +140,14 @@ func (r *RollingUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{RequeueAfter: v1alpha1.DefaultRequeueTime}, nil
 	}
 
+	// only reconcile maxParallel number of CRs at a time
+	if common.GetSyncMapLen(&r.AdmissionMap) >= r.maxParallel {
+		if _, present := r.AdmissionMap.Load(rollingUpgrade.NamespacedName()); !present {
+			r.Info("number of ongoing CRs has reached the max-parallel limit, requeuing", "name", rollingUpgrade.NamespacedName())
```
shreyas-badiger, line 146: @ZihanJiang96 did you get a chance to test this on a large cluster? By enforcing this limit, we might slow down upgrades significantly. 

Here's what we could do:
- Create a dev cluster with 100 IGs. (maybe 2 nodes in each.)
- Do a cluster upgrade with current version of upgrade-manager (with maxUnavailable set to 100%)
    - Monitor the memory & CPU consumption.
    - Measure how much time it took, overall.
- Do a cluster upgrade with your changes. (with maxUnavailable set to 100%)
    - Monitor the memory & CPU consumption.
    - Measure how much time it took, overall.

figo (approved):

ZihanJiang96:
the CI failed, but i don't have permission to rerun it

tekenstam (approved):
LGTM, pending fixes to CI pipeline.

controllers/rollingupgrade_controller.go
```diff
@@ -122,6 +131,22 @@ func (r *RollingUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if common.GetSyncMapLen(&r.AdmissionMap) == 0 {
+		r.Info("admission map is empty, fetching all running CRs")
+		var rollingUpgradeList v1alpha1.RollingUpgradeList
+		err := r.List(ctx, &rollingUpgradeList, client.MatchingLabels{v1alpha1.LabelKeyRollingUpgradeCurrentStatus: v1alpha1.StatusRunning})
+		if err != nil {
+			r.Error(err, "failed to fetch running CRs")
+			return ctrl.Result{}, err
+		}
+		for _, cr := range rollingUpgradeList.Items {
+			if cr.CurrentStatus() == v1alpha1.StatusRunning {
```
shreyas-badiger, line 144: We can probably skip this check if you are only listing the `Running` rolling upgrades.

shreyas-badiger:
@ZihanJiang96 nice work!

Just two concerns:
- Testing this on a cluster with multiple IGs. 
- The log messages for other IGs that are not admitted could become very noisy. (maybe skip printing them)

PR #394, opened 2023-10-19
Process drain-failure nodes at the end

ssheladiya:
**Problem:**
At present, if upgrade-manager comes across a node failure, it becomes stuck and fails to proceed with draining other nodes until the failed node is repaired. This may cause significant delays if the drain-failure node requires a considerable amount of time to fix

**Proposal:**
Upgrade-manager skip the drain-failed nodes and return to them when all the other nodes in the InstanceGroups are rotated

**Changes introduced by this PR:**
- Marks instances that failed to drain with `failed-drain` value
- Upgrade-manager skips `failed-drain` instances while selecting the target and moves on to other instances
- Once no healthy instances are left to rotate, it tries to rotate the `failed-drain` instances again


controllers/upgrade.go
```diff
@@ -479,6 +497,16 @@ func (r *RollingUpgradeContext) SelectTargets(scalingGroup *autoscaling.Group) [
 	return targets
 }
```
ZihanJiang96, line 499: no need to add this func.
we already had https://github.com/keikoproj/upgrade-manager/blob/84dd4a7d7bd7b0f86bf3969d0821bf920ef27510/controllers/common/utils.go#L25

controllers/upgrade.go
```diff
@@ -112,7 +117,14 @@ func (r *RollingUpgradeContext) RotateNodes() error {
 		return nil
 	}
 
-	rotationTargets := r.SelectTargets(scalingGroup)
+	rotationTargets := r.SelectTargets(scalingGroup, failedDrainInstances)
+
+	if len(rotationTargets) == 0 && len(failedDrainInstances) > 0 {
+		// If there are failed instances, but no rotation targets, then we need to requeue.
```
ZihanJiang96, line 123: don't quite understand the comment here.

based on the logic, it should be:
```suggestion
		// If there are failed instances, but no rotation targets, then select failed isntances anyway.
```

controllers/upgrade.go
```diff
@@ -418,7 +433,7 @@ func (r *RollingUpgradeContext) ReplaceNodeBatch(batch []*autoscaling.Instance)
 	return true, nil
 }
 
-func (r *RollingUpgradeContext) SelectTargets(scalingGroup *autoscaling.Group) []*autoscaling.Instance {
+func (r *RollingUpgradeContext) SelectTargets(scalingGroup *autoscaling.Group, failedDrainInstances []string) []*autoscaling.Instance {
```
ZihanJiang96, line 436:
```suggestion
func (r *RollingUpgradeContext) SelectTargets(scalingGroup *autoscaling.Group, excludedInstances []string) []*autoscaling.Instance {
```

controllers/upgrade.go
ssheladiya: Thanks for pointing out, I will use that

controllers/upgrade.go
ssheladiya: I agree, let me rephrase it

controllers/upgrade.go
ssheladiya: fixed in [4511ff5](https://github.com/keikoproj/upgrade-manager/pull/394/commits/4511ff524f3093469ef17f72515d6d26f946b46c)

controllers/upgrade.go
ssheladiya: fixed in [4511ff5](https://github.com/keikoproj/upgrade-manager/pull/394/commits/4511ff524f3093469ef17f72515d6d26f946b46c)

controllers/upgrade.go
ssheladiya: fixed in [1fab0cc](https://github.com/keikoproj/upgrade-manager/pull/394/commits/1fab0cc975e4f7af8b35f4e89a4244467683f4bc)

ZihanJiang96 (approved):

tekenstam (approved):

shreyas-badiger (approved):
LGTM

PR #405, opened 2023-11-09
Early cordon

shreyas-badiger:
Currently, upgrade-manager supports 2 different strategies:
**Eager mode** - Eagerly wait for replacement nodes and only then drain & terminate the previous instances. 
**Lazy mode** - Rotate (drain and terminate) the desired number of nodes without waiting for the replacement nodes.

In these two strategies, we cordon only the nodes that are in the current batch (batch size is determined by maxUnavailable mentioned in the RollingUpgrade CR. By default maxUnavailable=1) 

While the upgrade is in progress, the remaining older nodes that are not yet considered in the node-rotation batch, might have newer deployments / pods scheduled. 

These newly scheduled pods could have yet another restart when the underlying older nodes are considered for rotation. 
There is also an added time for draining these nodes with additional new pods.

With the approach in PR, we will cordon all the nodes in the respective IG when a rollingUpgrade CR is being processed. The newer pods will always scheduled on newer nodes when an upgrade is in progress.


controllers/providers/kubernetes/nodes.go
```diff
@@ -75,3 +76,33 @@ func (k *KubernetesClientSet) DrainNode(node *corev1.Node, PostDrainDelaySeconds
 	}
 	return nil
 }
+
+// CordonNode cordons a node.
+func (k *KubernetesClientSet) CordonUncordonNode(node *corev1.Node, client kubernetes.Interface) error {
```
vgunapati, line 81:
```suggestion
func (k *KubernetesClientSet) CordonNode(node *corev1.Node, client kubernetes.Interface) error {
```

controllers/providers/kubernetes/nodes.go
```diff
@@ -75,3 +76,33 @@ func (k *KubernetesClientSet) DrainNode(node *corev1.Node, PostDrainDelaySeconds
 	}
 	return nil
 }
+
+// CordonNode cordons a node.
```
vgunapati, line 80:
```suggestion
// CordonUncordonNode cordons a node.
```

controllers/upgrade.go
```diff
@@ -750,3 +758,51 @@ func (r *RollingUpgradeContext) ClusterBallooning(batchSize int) (bool, int) {
...
+				if node == nil {
+					r.Info("node object not found in clusterNodes, unable to early-cordon node", "instanceID", instance.InstanceId, "name", r.RollingUpgrade.NamespacedName())
+					continue
+				}
+				//Early cordon only the dirfted instances and not the instances that have same scaling-config as the scaling-group
+				if !r.IsInstanceDrifted(instance) {
+					break
+				}
+				r.Info("early cordoning node", "instanceID", instance.InstanceId, "name", r.RollingUpgrade.NamespacedName())
+				if err := r.Auth.CordonUncordonNode(node, r.Auth.Kubernetes, true); err != nil {
```
vgunapati, line 794: test Uncordon case

shreyas-badiger: Done.

shaoxt (approved):
LGTM

vgunapati (approved):

controllers/rollingupgrade_controller.go
```diff
@@ -212,13 +213,20 @@ func (r *RollingUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	}
 
 	// process node rotation
 	if err := rollupCtx.RotateNodes(); err != nil {
 		rollingUpgrade.SetCurrentStatus(v1alpha1.StatusError)
 		rollingUpgrade.SetLabel(v1alpha1.LabelKeyRollingUpgradeCurrentStatus, v1alpha1.StatusError)
 		common.SetMetricRollupFailed(rollingUpgrade.Name)
+
+		// try to uncordon all the cordoned nodes.
+		if _, err2 := rollupCtx.CordonUncordonAllNodes(false); err2 != nil {
```
ZihanJiang96, line 226: trying to understand why we do Uncordon here? If all old nodes remain in cordon state and there are pods pending, cluster autoscaler will launch new nodes, which is good.

shreyas-badiger: No. Think about cases where new nodes aren't coming up because of a faulty AMI. In such cases, we will have to uncordon the nodes or risk pod scaling / failure of new rollouts.

ZihanJiang96: Thanks!!!

PR #518, opened 2025-07-10
#7684 cordon nodes race condition

pritheeroy:
## Overview

This PR significantly improves the early cordoning feature by replacing AWS instance tags with Kubernetes-native node annotations and fixing race conditions and reliability issues.

## Describe the Problem

The current early cordoning implementation had some issues:
- Uses AWS instance tags instead of Kubernetes-native tracking
- Breaking on first non-drifted instance instead of processing all drifted instances
- No cleanup mechanism for cordoned nodes on failure
- Non robust error handling that fails entire operation on single node errors

### **Kubernetes-Native Tracking**
- Replace AWS instance tags with node annotation: `rollingupgrade.keikoproj.io/early-cordoned-by: upgrade-manager`
- Track which nodes were cordoned by upgrade-manager vs other processes
- Enable proper cleanup of only our cordoned nodes

### **Race Condition Fixes**
- Add `ensureClusterNodesMapPopulated()` to populate `ClusterNodesMap` before early cordoning
- Fix timing issue where NodeEventsHandler only captures nodes created after controller start

### **Reliability Improvements**
- Add retry mechanism with exponential backoff (100ms, 200ms, 300ms) for resource version conflicts
- Skip already cordoned nodes to avoid unnecessary API calls
- Skip instances in terminating states
- Process ALL drifted instances (fix `break` → `continue`)
- Continue processing nodes even if individual nodes fail
- Add cleanup logic on both success and failure paths

### **Error Handling**
- Graceful handling of individual node failures
- Rollback cordoning if annotation fails
- Comprehensive error reporting with node-level details

### **Test Environment:**
- **Cluster**: 30-node EKS cluster
- **Instance Types**: m6a.xlarge → c6a.xlarge upgrade
- **ASG**: `dev-proyballave-2-instance-manager-nodes`
- **Strategy**: `uniformAcrossAzUpdate` with 5% maxUnavailable

### **Test Results:**
- **All 30 nodes successfully cordoned**
-**Zero errors** during cordoning process  
- **All nodes properly annotated** with `rollingupgrade.keikoproj.io/early-cordoned-by: upgrade-manager`
- **Skip logic working**: Subsequent reconciliations show "node already cordoned by upgrade-manager, skipping"
- **Node rotation successful**: Instances properly drained and terminated
- **Resource version conflicts handled**: Retry mechanism with exponential backoff working


controllers/upgrade.go
```diff
@@ -777,57 +840,167 @@ func (r *RollingUpgradeContext) ClusterBallooning(batchSize int) (bool, int) {
...
-			if !common.ContainsEqualFold(awsprovider.TerminatingInstanceStates, aws.StringValue(instance.LifecycleState)) {
-				node := kubeprovider.SelectNodeByInstanceID(*instance.InstanceId, r.Cloud.ClusterNodes)
-				if node == nil {
-					r.Info("node object not found in clusterNodes, unable to early-cordon node", "instanceID", instance.InstanceId, "name", r.RollingUpgrade.NamespacedName())
+		if cordonNode {
+			// Skip already cordoned nodes to avoid unnecessary API calls
+			if node.Spec.Unschedulable {
+				// Check if it was cordoned by upgrade-manager
+				if node.Annotations != nil && node.Annotations[EarlyCordonAnnotationKey] == EarlyCordonAnnotationValue {
+					r.Info("node already cordoned by upgrade-manager, skipping", "instanceID", instance.InstanceId, "nodeName", node.Name, "name", r.RollingUpgrade.NamespacedName())
```
shreyas-badiger, line 879: Remove the "... by upgrade-manager" from the log. Node could be cordoned by other entities too - cluster-autoscaler, lifecycle-manager, node-reaper.

controllers/upgrade.go
```diff
@@ -777,57 +840,167 @@ func (r *RollingUpgradeContext) ClusterBallooning(batchSize int) (bool, int) {
...
+
+		// Update the node with the annotation
+		if _, err := r.Auth.Kubernetes.CoreV1().Nodes().Update(context.Background(), freshNode, metav1.UpdateOptions{}); err != nil {
+			if attempt < maxRetries && strings.Contains(err.Error(), "the object has been modified") {
+				r.Info("resource version conflict, retrying annotation", "nodeName", freshNode.Name, "attempt", attempt, "maxRetries", maxRetries)
+				time.Sleep(time.Duration(attempt) * 100 * time.Millisecond) // Exponential backoff
+				continue
+			}
+			// If annotation fails after all retries, try to uncordon to avoid leaving node in inconsistent state
+			if uncordonErr := r.Auth.CordonUncordonNode(freshNode, r.Auth.Kubernetes, false); uncordonErr != nil {
```
shreyas-badiger, line 955: Probably you can handle al the uncordon scenarios in uncordonAndRemoveAnnotation() function. Just add a a check in the uncordonAndRemoveAnnotation() function to ignore annotations if they are not present.

controllers/upgrade.go
```diff
@@ -777,57 +840,167 @@ func (r *RollingUpgradeContext) ClusterBallooning(batchSize int) (bool, int) {
...
+	}
+
+	// Retry annotation update with fresh node objects to handle resource version conflicts
+	maxRetries := 3
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		// Get fresh node object to avoid resource version conflicts
+		freshNode, err := r.Auth.Kubernetes.CoreV1().Nodes().Get(context.Background(), node.Name, metav1.GetOptions{})
+		if err != nil {
+			// If we can't get the fresh node, try to uncordon to avoid inconsistent state
+			if uncordonErr := r.Auth.CordonUncordonNode(node, r.Auth.Kubernetes, false); uncordonErr != nil {
```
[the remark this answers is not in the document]
pritheeroy: We uncordon the node since we can't annotate it to track our cordon operation, which could cause an orphaned node (then causing the node to be cordoned forever) during cleanup if it doesn't have the cordoned annotation.

controllers/upgrade.go
pritheeroy: That is a good suggestion, it would make uncordonAndRemoveAnnotation handle all uncordoning scenarios

controllers/upgrade.go
```diff
@@ -777,57 +840,193 @@ func (r *RollingUpgradeContext) ClusterBallooning(batchSize int) (bool, int) {
...
+	var processedCount, errorCount int
+
+	for _, instance := range scalingGroup.Instances {
+		// Skip terminating instances
+		if common.ContainsEqualFold(awsprovider.TerminatingInstanceStates, aws.StringValue(instance.LifecycleState)) {
+			continue
+		}
+
+		// Skip non-drifted instances
+		if !r.IsInstanceDrifted(instance) {
```
shreyas-badiger, line 864: Ideally, non drifted instance should never be considered for earlyCordon. 

But in case a non-drifted instance was early cordoned (when the launch template is switched back and forth. instance-manager could do this, if the latest launch template has issues.) after checking for the annotation, we should be uncordoning it.

tekenstam:
@pritheeroy When working on keikoproj repos, please "sign-off" your commits so that the DCO check will pass. You can do this by adding the `-s` option to `git commit -s`.

tekenstam:
Is this potentially a breaking change since it moves from AWS instance tags to Kubernetes annotations? @shreyas-badiger Anything we should do (AWS tags vs K8s annotations) when rolling this change out?

controllers/upgrade.go
pritheeroy: Sounds good, I added an extra conditional to ensure that we do not consider non-drifted instances when cordonNode is true.

For the case you mentioned, I believe we handle that here:
https://github.com/keikoproj/upgrade-manager/pull/518/files#diff-077ba5eb59266ac27713b3442231daf3287b93fa4b6683c7a58aea43c0e73f12R890-R900

